PR #3, opened 2023-09-05
@W-14071095@: Unit tests now collect code coverage.

jfeingold35:


.github/workflows/publishVSCode.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 'lts/*'
-      - run: yarn install --frozen-lockfile && yarn global add @vscode/vsce
+      - run: yarn install --frozen-lockfile
```
jfeingold35, line 79: Since `@vscode/vsce` is a dependency now, no need to do the global add.

src/test/suite/extension.test.ts
```diff
@@ -6,18 +6,84 @@
...
 // as well as import your extension to test it
 import * as vscode from 'vscode';
-// import * as myExtension from '../../extension';
 
 suite('Extension Test Suite', () => {
 	vscode.window.showInformationMessage('Start all tests.');
+	// Note: __dirname is used here because it's consistent across file systems.
+	const codeFixturesPath: string = path.resolve(__dirname, '..', '..', '..', 'code-fixtures');
+
+	suite('E2E', () => {
```
jfeingold35, line 24: Adding integration/E2E tests is technically part of another story, but the current structure of the extension required me to add at least one E2E test to get coverage numbers to where they needed to be.

src/test/suite/extension.test.ts
```diff
@@ -6,18 +6,84 @@
...
+
+			// At present, we expect only violations for PMD's `ApexDoc` rule.
+			for (const diagnostic of diagnostics) {
+				expect(diagnostic.source).to.equal('pmd via Code Analyzer', 'Wrong source');
+				expect(diagnostic.code).to.have.property('value', 'ApexDoc', 'Wrong rule violated');
+			}
+		});
+
+		test('sfca.runOnSelected', async function() {
+			// TODO: Add actual tests for `runOnSelected`.
```
jfeingold35, line 78: If we still want to do the other story for adding further integration/E2E tests, they can be added in these TODOs.

src/test/suite/targeting.test.ts
```diff
@@ -136,12 +142,13 @@ suite('targeting.ts', () => {
...
-            for (let i = 0; i < tabs.length; i++) {
-                await vscode.window.tabGroups.close(tabs[i]);
-            }
+			// ===== SETUP =====
+			// Simulate no window being open in the editor.
+			// NOTE: We need to use a stub here instead of/in addition to directly closing
+			//       windows, because sometimes the test context can have a weird "phantom window"
+			//       even if you close anything. This seems to happen when the test instance loses
+			//       focus.
+			const windowStub = Sinon.stub(vscode.window, 'activeTextEditor').value(undefined);
```
jfeingold35, line 151: I know this stub seems redundant with line 26. The thing is, there's an odd behavior where if you're running the tests locally and you click away from the window, then the `closeAllEditors` command causes a weird phantom editor to open that doesn't open if the window is still in focus. I could find no information about this phantom editor online.
So in order to allow you to tab away while tests are running, I'm just stubbing out the property as `undefined` to work around that.

ganeshnt (approved):

smotwani11 (approved):

PR #27, opened 2023-12-01
@W-14219242@: Fixing behavioral issues with telemetry.

jfeingold35:


src/lib/core-extension-service.ts
```diff
@@ -79,14 +79,34 @@ export class CoreExtensionService {
 	 *
 	 * @returns The {@link TelemetryService} object exported by the Core Extension if available, else null.
 	 */
-	public static getTelemetryService(): TelemetryService|null {
+	public static _getTelemetryService(): CoreTelemetryService|null {
```
jfeingold35, line 82: The leading underscore is a JS social convention indicating that the method should be treated by developers as private even though it's technically public.

src/lib/core-extension-service.ts
```diff
@@ -79,14 +79,34 @@ export class CoreExtensionService {
...
 		if (CoreExtensionService.initialized) {
 			return CoreExtensionService.telemetryService;
 		}
 		throw new Error(messages.error.coreExtensionServiceUninitialized);
 	}
 }
 
+export class TelemetryService {
```
jfeingold35, line 90: Moving `TelemetryService` into `core-extension-service.ts` allowed me to make it so all telemetry events have to go through this class instead of calling `CoreExtensionService._getTelemetryService()` directly.

src/lib/telemetry.ts
```diff
@@ -1,24 +0,0 @@
...
- */
-import { CoreExtensionService, Properties } from './core-extension-service';
-
-/**
- * Helper class for managing access to the core telemetry service.
- */
-export class TelemetryService {
-	public static sendCommandEvent(key: string, data: Properties): void {
-		const coreTelemetryService = CoreExtensionService.getTelemetryService();
-		coreTelemetryService.sendCommandEvent(key, undefined, data);
```
jfeingold35, line 15: This needed to be `coreTelemetryService?.sendCommandEvent`, like how it's done in `sendException`. That error has been corrected in the version that now exists in `core-extension-service.ts`.

PR #48, opened 2024-02-27
@W-15093436@ trigger scan on save when a file is saved

jag-j:


src/extension.ts
```diff
@@ -58,6 +58,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	const outputChannel: vscode.LogOutputChannel = vscode.window.createOutputChannel('sfca', {log: true});
 	outputChannel.clear();
 	outputChannel.show();
+	outputChannel.appendLine(`Extension sfdx-code-analyzer-vscode activated.`);
```
jfeingold35, line 61: Question: Should this log line be at the end of the method?

src/lib/settings.ts
```diff
@@ -29,4 +29,8 @@ export class SettingsManager {
     public static getGraphEngineJvmArgs(): string {
         return vscode.workspace.getConfiguration('codeAnalyzer.graphEngine').get('jvmArgs');
     }
+
+    public static getAnalyzeOnSave(): boolean {
+        return vscode.workspace.getConfiguration('codeAnalyzer.analyze-on-save').get('enabled');
```
jfeingold35, line 34: You'll need to add a new property for this in the `contributes.configuration.properties` section of `package.json`, I believe. You'll also need to clear the property's label and help text with Teresa.

jfeingold35 (commented):
Question: is it possible to add test coverage for the new feature?

src/lib/settings.ts
jag-j: I sent a separate PR for that - https://github.com/forcedotcom/sfdx-code-analyzer-vscode/pull/46 and got it merged. I didn't get a review from @teresa-allen-sfdc. I will follow up on that.

jag-j:
@jfeingold35 - Unfortunately there isn't a way to add tests for activation events AFAIK. I looked at some examples on other extensions and they don't seem to have either.

jfeingold35:
@jag-j , can you add a test for the scan being triggered after a file is saved?

stephen-carter-at-sf:
Food for thought regarding:
>  Unfortunately there isn't a way to add tests for activation events AFAIK. I looked at some examples on other extensions and they don't seem to have either.

There is always a way to test... but it may require a number of layers of indirection. The trade off is that the code may become unnecessarily complex if we add in a ton of interfaces to achieve indirection and testability. But in this case it seems that if we added in the interfaces for indirection, we would simply be testing that our interfaces were called and wouldn't provide much value other than code coverage purposes. So this mainly depends on how high we want our code coverage numbers. But in terms of functionality testing... doesn't help much if the test code is highly coupled to the source code due to the nature of the feature we are testing.

PR #52, opened 2024-03-21
@W-15253475@ remove cli dependency in the extensions

jag-j:


src/lib/display.ts
```diff
@@ -0,0 +1,67 @@
+import { Display, Displayable } from '@salesforce/sfdx-scanner/lib/lib/Display';
+import {Spinner, Ux} from "@salesforce/sf-plugins-core";
+import {AnyJson} from "@salesforce/ts-types";
+
+
+export class VSCodeDisplay implements Display {
```
jag-j, line 6: This is completely a dummy implementation of Display. I do not see a need to change anything here since the display is fully handled on the VS Code side. For 5.x, we should revisit this. Do we need to pass in a Display object to the "core" package? Or is there any way for CLI to handle all the display based on the events it receives from core? I'm curious how source-deploy-retrieve package does it now.

src/lib/scanner.ts
```diff
@@ -137,77 +132,25 @@ export class ScanRunner {
...
-     * @param executionResult The results from a scan
-     * @returns The HTML-formatted scan results, or an empty string.
-     * @throws If {@code executionResult.result} is not a string.
-     * @throws If {@code executionResult.warnings} contains any warnings about methods not being found.
-     * @throws if {@code executionResult.status} is non-zero.
+        /**
+     * Uses the provided arguments to run a Salesforce Code Analyzer command.
+     * @param args The arguments to be supplied
      */
-    private processDfaResults(executionResult: ExecutionResult): string {
```
jag-j, line 148: Didn't see a need for this method now that we are getting a HTML directly from the runAction.

src/lib/types.d.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+export type Inputs = {
```
jag-j, line 8: Ideally I wanted to use the "Inputs" declaration from sfdx-scanner's types.d.ts file. But I can't access that file's declarations from vs code. That's also the reason for the compilation errors I ran into locally and in the build.

package.json
```diff
@@ -33,6 +33,7 @@
         "SFCA"
     ],
     "dependencies": {
+        "@salesforce/sfdx-scanner": "3.22.0",
```
stephen-carter-at-sf, line 36: If we do end up shipping this... then we'll want to swap this out for 4.0.0 immediately after we release the 4.0.0 cli.

src/extension.ts
```diff
@@ -75,6 +76,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 		});
 	});
 	outputChannel.appendLine(`Registered command as part of sfdx-code-analyzer-vscode activation.`);
+	registerAll();
```
stephen-carter-at-sf, line 79: I see this is inside the activate function. Does this run every time the scanner's action is invoked - or does vs code keep track of things in memory? I ask because registerAll needs to have been invoked to make the unfortunately global state ready for the action to perform properly. If not, then we may need to move this registerAll as the first step of the run method in the scanner.ts file.

src/lib/display.ts
stephen-carter-at-sf: Did the old vs code extension implementation give real time updates from when the scanner was running? And did it forward error/warning messages to the user somehow? If so, then I don't think we can have a 100% dummy display because the display is wired up the the events and is wired up to push errors to.

src/lib/scanner.ts
```diff
@@ -39,53 +50,46 @@ export class ScanRunner {
...
-        const args: string[] = this.createDfaArgArray(targets, projectDir);
+        const args: Inputs = this.createDfaArgArray(targets, projectDir);
 
         // Invoke the scanner.
-        const executionResult: ExecutionResult = await this.invokeAnalyzer(args);
+        const executionResult: AnyJson = await this.invokeAnalyzerWithoutCli(args);
 
         // Process the results.
-        return this.processDfaResults(executionResult);
+        return executionResult as string;
```
stephen-carter-at-sf, line 59: Is the AnyJson returned always a string? If not then this cast will fail in non-string cases. We'll need to double check.

src/lib/scanner.ts
```diff
@@ -21,13 +30,15 @@ export class ScanRunner {
      */
     public async run(targets: string[]): Promise<RuleResult[]> {
         // Create the arg array.
-        const args: string[] = await this.createPathlessArgArray(targets);
+        const args: Inputs = this.createPathlessArgArray(targets);
 
         // Invoke the scanner.
-        const executionResult: ExecutionResult = await this.invokeAnalyzer(args);
+        // const executionResult = await this.invokeAnalyzer(args);
```
stephen-carter-at-sf, line 36: is there a reason why there is commented out code still?

src/lib/scanner.ts
```diff
@@ -21,13 +30,15 @@ export class ScanRunner {
...
+        const executionResult: AnyJson = await this.invokeAnalyzerWithoutCli(args);
 
         // Process the results.
-        return this.processPathlessResults(executionResult);
+        // return this.processPathlessResults(executionResult); 
+        return executionResult as RuleResult[];
```
stephen-carter-at-sf, line 41: It seems odd that this run method returns a RuleResult array but the runDfa method returns a string. Why are they different?

And how is it possible that the invokeAnalyzerWithoutCli output can be casted to a RuleResult array and the other to a string? That seems odd to me.

src/lib/scanner.ts
```diff
@@ -39,53 +50,46 @@ export class ScanRunner {
...
-            //       elements of logging, most notably the informative progress reporting provided
-            //       by the engine's spinner. This was deemed an acceptable trade-off during initial
-            //       implementation, but we may wish to rethink this in the future as we polish things.
-            `--json`
-        ];
+    private createDfaArgArray(targets: string[], projectDir: string): Inputs {
+        const args: Inputs = {
+            'targets': `${targets.join(',')}`,
+            'projectdir': [projectDir],
+            'format': `html`,
```
stephen-carter-at-sf, line 71: Why do we format as html for runDfa but json for run? That seems odd.

src/lib/scanner.ts
```diff
@@ -98,20 +102,11 @@ export class ScanRunner {
...
-            '--engine', 'pmd,retire-js',
-            '--json'
-        ];
-        const customPmdConfig: string = SettingsManager.getPmdCustomConfigFile();
-        // If there's a non-null, non-empty PMD config file specified, use it.
-        if (customPmdConfig && customPmdConfig.length > 0) {
-            if (!(await exists(customPmdConfig))) {
-                throw new Error(messages.error.pmdConfigNotFoundGenerator(customPmdConfig));
-            }
-            args.push('--pmdconfig', customPmdConfig);
```
stephen-carter-at-sf, line 114: What happened to this customPmdConfig? Why did it get dropped?

src/lib/types.d.ts
stephen-carter-at-sf: Hmmm... that's odd. Maybe its because it is named types.d.ts instead of types.ts (not sure why it was named that way). But theoretically we should be able to export he "Inputs"... but this is fine to.

stephen-carter-at-sf (changes requested):
No tests?

I don't think this implementation will match what vs code extension does today with real time updates of the output. Also i'm confused as to how you are casting the AnyJson to a string in one case but a RuleResult array in another case. That doesn't seem correct to me.

src/lib/scanner.ts
stephen-carter-at-sf, line 148: I'm surprised that we can remove this method. I would have thought that we would be processing the same text (whether it was retrieved from the display or from json output) just as before.

src/lib/scanner.ts
```diff
@@ -21,13 +30,15 @@ export class ScanRunner {
      */
     public async run(targets: string[]): Promise<RuleResult[]> {
         // Create the arg array.
-        const args: string[] = await this.createPathlessArgArray(targets);
+        const args: Inputs = this.createPathlessArgArray(targets);
 
         // Invoke the scanner.
-        const executionResult: ExecutionResult = await this.invokeAnalyzer(args);
```
stephen-carter-at-sf, line 27: If we are no longer calling invokeAnalyzer... then how come I don't see it being deleted in this review?

src/extension.ts
jfeingold35: `activate` runs when the extension is activated. Currently, the events for that are "a project with a `sfdx-project.json` file being opened" and "one of the commands being executed for the first time (if the extension isn't already active)".

src/lib/display.ts
jfeingold35: The extension has a progress bar that _very_ loosely guesses where you are in the job.

src/lib/scanner.ts
jfeingold35: Because DFA violations have both a source and a sink, whereas non-DFA violations just have one location, it wasn't immediately clear how to actually display information about DFA violations inline.
So we just have the scanner return results in its existing HTML format, and then we display that content in a new tab.

src/lib/types.d.ts
jag-j: This is now taken care of.

package.json
jag-j: Updated this to 4.0.0.

src/extension.ts
jag-j: Josh is correct. As we discussed last week, this is now activated once a project with a sfdx-project.json file (a valid DX project) is loaded. Resolving this comment. Please re-open if there are any other concerns.

src/lib/scanner.ts
jag-j: Yes, AnyJson is always returned a string for DFA case. That's because we request the output to be in HTML format (as opposed to JSON for the non-DFA case). It'll be good to revisit the output type while designing the core package to accommodate this. (Leaving this comment open for visibility).

src/lib/scanner.ts
jag-j: This can be removed. I've removed it now.

src/lib/scanner.ts
jag-j: This was deleted by mistake. I added this back in.

src/lib/scanner.ts
jag-j: Deleted.

src/lib/scanner.ts
jag-j: I need @jfeingold35's inputs here. But I think the executionResult coming out of CLI stdout had other parameters and we needed this to parse and just get to the HTML. Based on my testing, by directly calling the runDfaAction, we only get the plain HTML and I don't see any issues with simply casting than to a string and showing it to the users. Let me know if I'm missing anything.

jag-j:
> c

I couldn't run the tests because of the index.d.ts dependency. This is now taken care and I've updated the tests.

src/lib/scanner.ts
```diff
@@ -98,116 +104,63 @@ export class ScanRunner {
...
-        return new Promise((res) => {
-            const cp = cspawn.spawn('sf', args);
-
-            let stdout = '';
-
-            cp.stdout.on('data', data => {
-                stdout += data;
-            });
+    private async invokeAnalyzer(args: Inputs): Promise<AnyJson> {
+        const sfVersion = '0.0.5';
```
jag-j, line 128: @jfeingold35 @stephen-carter-at-sf  - Where is this version used? I could instead read this from package.json or have it set in the settings. But I'm still trying to understand the significance of this.

stephen-carter-at-sf: Yeah this I really dislike. The "sf" cli version is something that is added to the RunOptions (see line 86 of InputProcessor) that we ultimately send off in our telemetry calls (see line 147 and 182 of DefaultRuleManager) as part of our telemetry data.

That is, our rule manager assumes that we are running from the CLI which is why this artifact still exists. You can probably just put in x.x.x if you want - again it's only for telemetry data.

src/lib/scanner.ts
jag-j: @stephen-carter-at-sf - Checked on the warning and error cases and we are good. The only case that was being handled was related to "no method found" and there isn't a possibility of this occurring via VS Code. I ran through this scenario with @jfeingold35 offline and he agrees on this approach.

src/lib/scanner.ts
jag-j: Keeping it as 0.5.0 for now.

src/test/suite/scanner.test.ts
```diff
@@ -271,199 +242,12 @@ suite('ScanRunner', () => {
...
                 // Verify that the right arguments were created.
-                expect(args).to.have.lengthOf(12, 'Wrong number of args');
-                expect(args[10]).to.equal('--sfgejvmargs', 'Wrong arg');
-                expect(args[11]).to.equal(jvmArgs, 'Wrong arg');
+                expect(args['sfgejvmargs']).to.equal(jvmArgs, 'Wrong arg');
             });
         });
     });
-
-    suite('#processPathlessResults()', () => {
```
stephen-carter-at-sf, line 285: Why did all these tests get deleted? Is this why the code coverage numbers dropped?

src/lib/scanner.ts
jag-j: Done.

src/lib/display.ts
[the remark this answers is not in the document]
jag-j: Good point. Done.

src/lib/types.d.ts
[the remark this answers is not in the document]
jag-j: My bad. Not sure how I missed this. Meant to delete it as soon as the CLI was released.

src/test/suite/scanner.test.ts
jag-j: That's because the methods they were testing were also deleted. Those methods were the ones parsing the stdout.
The test coverage drop is however unrelated and it's because of the noop display methods.

src/lib/scanner.ts
```diff
@@ -5,29 +5,42 @@
...
-        const args: string[] = await this.createPathlessArgArray(targets);
+        const args: Inputs = await this.createPathlessArgArray(targets);
 
         // Invoke the scanner.
-        const executionResult: ExecutionResult = await this.invokeAnalyzer(args);
+        const executionResult: AnyJson = await this.invokeAnalyzer(args, false);
 
         // Process the results.
-        return this.processPathlessResults(executionResult);
+        // return this.processPathlessResults(executionResult);
```
stephen-carter-at-sf, line 42: Can we remove this line of commented out code?

src/lib/display.ts
```diff
@@ -0,0 +1,74 @@
...
+    public displayStyledObject(obj: AnyJson): void {
+	}
+
+    public getInfo(): string[] {
+        return this.stdout;
+    }
+
+    public getErrorsAndWarnings(): string [] {
+        return this.stderr;
+    }
```
stephen-carter-at-sf, line 73: I see you switched back to VSCodeDisplay (instead of NoOpDisplay) and you added these getters. I only see this being used in a unit test file... but never used in the source code.

What is the purpose of this? Is it just for code coverage purposes? If so, then why wouldn't NoOpDisplay still work?

Or is there a plan to use these getters later in the source code to handle error/warning cases better?

src/lib/scanner.ts
```diff
@@ -98,116 +104,46 @@ export class ScanRunner {
...
+        const display:Display = new VSCodeDisplay();
+        const inputProcessor: InputProcessor = new InputProcessorImpl(sfVersion, display);
+        const ruleFilterFactory: RuleFilterFactory = new RuleFilterFactoryImpl();
+        const resultsProcessorFactory: ResultsProcessorFactory = new ResultsProcessorFactoryImpl();
+        if (isDfa) {
+            const engineOptionsFactory: EngineOptionsFactory = new RunDfaEngineOptionsFactory(inputProcessor);
+            const runAction:RunDfaAction = new RunDfaAction(logger, display, inputProcessor, ruleFilterFactory, engineOptionsFactory,
+                resultsProcessorFactory);
+            await runAction.validateInputs(args);
+            return runAction.run(args);
```
stephen-carter-at-sf, line 140: At some point, since you are gathering the errors and warnings with your latest VSCodeDisplay - it might be worth processing them here to be put somewhere inside of vscode's output window (or maybe there is an error window or something they can go into... i'm not sure what the standards are) before returning.

Ditto regarding line 146.

src/lib/scanner.ts
```diff
@@ -98,116 +104,46 @@ export class ScanRunner {
...
+                    } else {
```
stephen-carter-at-sf, line 141: The whitespace seems off here compared to line 135.

src/test/suite/scanner.test.ts
```diff
@@ -42,14 +44,11 @@ suite('ScanRunner', () => {
...
                 // Verify that the right arguments were created.
-                expect(args).to.have.lengthOf(7, 'Wrong number of args');
-                expect(args[0]).to.equal('scanner', 'Wrong arg');
-				expect(args[1]).to.equal('run', 'Wrong arg');
-                expect(args[2]).to.equal('--target', 'Wrong arg');
-                expect(args[3]).to.equal(targets.join(','), 'Wrong arg');
-                expect(args[4]).to.equal('--engine', 'Wrong arg');
-                expect(args[5]).to.equal('pmd,retire-js', 'Wrong arg');
-                expect(args[6]).to.equal('--json', 'Wrong arg');
+                // expect(args).to.have.lengthOf(3, 'Wrong number of args');
```
stephen-carter-at-sf, line 47: Did you mean to have commented out code here? Maybe just remove this line if not needed.

src/test/suite/scanner.test.ts
stephen-carter-at-sf: Were these tests just testing implementation detail or where they testing behavior? Ideally tests should lock in requirements and contracts... not implementation detail. So if the requirement is that we somehow handle errors/warnings a specific way - then we should keep that tested somehow.

I don't know exactly what these tests were for but they see things like... 
"Returns violations found during successful scan"
"Escalates method-not-found warning to error"
Are these no longer a part of the overall contract?

If they are, then tests should be modified instead of deleted (ideally to not be implementation specific if possible).

If they were just testing implementation detail... then fine these can go away if they truly didn't lock in some requirement or contract.

src/lib/display.ts
```diff
@@ -0,0 +1,67 @@
+import { Display, Displayable } from '@salesforce/sfdx-scanner/lib/lib/Display';
+import {Spinner, Ux} from "@salesforce/sf-plugins-core";
+import {AnyJson} from "@salesforce/ts-types";
+
+
+export class VSCodeDisplay implements Display {
+    private readonly displayable: Displayable;
+	private readonly spinner: Spinner;
+	private readonly isVerboseSet: boolean;
```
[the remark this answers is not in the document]
jag-j: @stephen-carter-at-sf - I have been going back and forth on this. There is still some implementation now and have added tests to please the code-coverage. But don't see any value in actually using the display. Though that would have been the simpler approach as you had suggested, there is value in consuming the results from the runAction to make us move closer to how it would be once we have the core package in place. Anyways, I have removed all the unused and added some new.

stephen-carter-at-sf (dismissed):
Approved mending removing commented out code at minimum.

src/test/suite/scanner.test.ts
jag-j: Done

src/lib/display.ts
jag-j: As discussed in another thread, these could go into the output window of the extension. But right now, this adds no value because there is no extra information here.

src/lib/scanner.ts
jag-j: Fixed

stephen-carter-at-sf (approved):

Issue #43, opened 2024-01-19
[BUG] linter links are not redirecting correctly to browser.  


TWood-IRL:
**Describe the bug**

When clicking link in vscode redirecting incorrectly. 
<img width="932" alt="image" src="https://github.com/forcedotcom/sfdx-code-analyzer-vscode/assets/10601890/b853bba4-cfe7-4ee2-b799-549993f3abb5">

Clicking link is redirecting within vs code. 
<img width="950" alt="image" src="https://github.com/forcedotcom/sfdx-code-analyzer-vscode/assets/10601890/21729303-f6cb-485e-bcc4-8733fa912e5e">

**To Reproduce**
> Open apex class file. 
> SF: Scan current file with code analyzer 
> View issues 
> Click link to doc 
> Doc link redirects incorrectly. 

**Expected behavior**

Should prompt to open link. (Taken from Apex PMD extension) 
<img width="277" alt="image" src="https://github.com/forcedotcom/sfdx-code-analyzer-vscode/assets/10601890/5b4ed565-304f-4c22-9d1c-90c0e2878c10">

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macOS - 14.2.1 (23C71)
 -  VSCode: Version: 1.85.1 (
 -  Salesforce Code Analyzer - salesforce.sfdx-code-analyzer-vscode - Version | 0.4.0

**Additional context**

"Workaround": Have you found any ways to sidestep the problem?

"Urgency": How severe is this problem? "Business stopping" vs "Not business-stopping".

PR #67, opened 2024-05-02
FIX (Extension) @W-15478761@: Fix rule violation url links

stephen-carter-at-sf:
With TDD approach, first I updated a test with an addition assertion to catch the bug. Then I updated the source code and made the test pass to confirm the bug fix.

I also manually installed and confirmed all works now.


stephen-carter-at-sf:
Fixes #43

jag-j (approved):

jag-j (approved):

PR #69, opened 2024-05-08
FIX (Extension) @W-15478740@: Limit scan on open operation to select types of files

jag-j:


src/extension.ts
```diff
@@ -279,3 +282,9 @@ async function summarizeResultsAsToast(targets: string[], results: RuleResult[])
 export function deactivate() {
 	TelemetryService.dispose();
 }
+
+export function _isValidFileForAnalysis(documentUri: vscode.Uri) {
+	const allowedFileTypes:string[] = ['.cls', '.js', '.apex', '.trigger', '.ts'];
+	return allowedFileTypes.includes(path.extname(documentUri.path));
```
stephen-carter-at-sf, line 288: Do we care about case insensitivity when validating file extensions?

jag-j: If someone creates apex class or any of these files using the extension, I'm pretty these are going to be in lower case. I have seen similar code in vscode extension where the checks are made only for lower case file. I think it is good as is.

stephen-carter-at-sf: Fair enough. Approved.

stephen-carter-at-sf (approved):

PR #70, opened 2024-05-09
NEW (Extension) @W-14001840@: Add ability to cancel graph engine runs

jag-j:


src/lib/messages.ts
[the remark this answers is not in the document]
jag-j: Yes, I am working with @jshackell-sfdc on those.

src/lib/messages.ts
```diff
@@ -25,8 +25,13 @@ export const messages = {
     graphEngine: {
         noViolationsFound: "Scan was completed. No violations found.",
```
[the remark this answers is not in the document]
jag-j: Yes, this is still used when there are no violations found. I tried this with some empty methods and made sure this is thrown.

src/extension.ts
```diff
@@ -204,41 +239,40 @@ export async function _runAndDisplayPathless(selections: vscode.Uri[], runInfo:
...
+					'dfaResults',
+					messages.graphEngine.resultsTab,
+					vscode.ViewColumn.Two,
+					{
+						enableScripts: true
+					}
+				);
+				panel.webview.html = results;
+			} else {
+				cancelToken.cancel();
```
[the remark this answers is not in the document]
jag-j: This is a workaround to make the progress window to disappear. The way progress window works is that there is no cancel API and instead, you use the progress information to automatically cancel the window. Since we don't get such information from graph engine, the workaround is to create a cancel token and call that.

src/test/suite/extension.test.ts
```diff
@@ -284,6 +284,111 @@ suite('Extension Test Suite', () => {
...
+			this.timeout(10000);
+			// Activate the extension.
+			context = await ext.activate();
+		});
+
+		teardown(async () => {
+			await context.globalState.update(Constants.GLOBAL_DFA_PROCESS, undefined);
+		});
+
+		test('Returns true when no existing DFA process detected', async() => {
```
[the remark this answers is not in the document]
jag-j: My bad! I was playing around with these two tests to see if it changes the code coverage number and missed to revert those changes. I have updated them now.

src/test/suite/extension.test.ts
```diff
@@ -284,6 +284,111 @@ suite('Extension Test Suite', () => {
...
+			await context.globalState.update(Constants.GLOBAL_DFA_PROCESS, 1234);
+			_shouldProceedWithDfaRun(context, outputChannel).then( (result) => {
```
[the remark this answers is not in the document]
jag-j: You are correct. It was supposed to test for the confirmation message pop up. I have taken care of it now.

src/extension.ts
```diff
@@ -78,14 +81,29 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	outputChannel.appendLine(`Registered command as part of sfdx-code-analyzer-vscode activation.`);
 	registerScanOnSave(outputChannel);
 	registerScanOnOpen(outputChannel);
+	await _stopExistingDfaRun(context, outputChannel, false);
```
stephen-carter-at-sf, line 84: This feels out of place. Why are we calling to the "stop existing dfa run" method inside of the activate method?

src/extension.ts
```diff
@@ -94,10 +112,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	return Promise.resolve(context);
 }
 
+export async function _stopExistingDfaRun(context: vscode.ExtensionContext, outputChannel: vscode.LogOutputChannel, verbose: boolean) {
```
stephen-carter-at-sf, line 115: Boolean arguments are typical code smells. 

I see that there is only one place where you are passing in false... in that case - I wonder if you should really be calling this method or if there should be a separate helper method for that case. Either in the other comment, or right here... can you explain why you are calling the stop existing dfa run method during activation? Is it simply to call through to 
context.globalState.update(Constants.GLOBAL_DFA_PROCESS, undefined);

If so, can we either just put that in the activate step... or put what you need for that in a separate method and try to keep each method having a single responsibility?

src/extension.ts
```diff
@@ -94,10 +112,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+	const pid = context.globalState.get(Constants.GLOBAL_DFA_PROCESS);
```
stephen-carter-at-sf, line 116: Do you know what happens if you have 2 vs code windows open? Do they share global state?

src/extension.ts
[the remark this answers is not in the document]
jfeingold35: Okay, but why did the `showInformationMessage` call get removed? Do we not need it anymore?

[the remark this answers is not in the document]
jag-j: It is not removed. It is now called as part of the cancel token invocation. See line number 99.

src/extension.ts
jag-j: Synced up with @stephen-carter-at-sf offline and we decide to keep this simple for now. The cache would be at the workspace level and we let people run multiple graph engines from multiple vscode windows.

src/extension.ts
jag-j: We allow this now since I switched the cache to be at the workspace level.

src/extension.ts
```diff
@@ -94,10 +111,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	return Promise.resolve(context);
 }
 
+export async function _stopExistingDfaRun(context: vscode.ExtensionContext, outputChannel: vscode.LogOutputChannel) {
```
jfeingold35, line 114: As a matter of style, typescript functions should have return types declared, so this should be `: Promise<void>`

src/extension.ts
```diff
@@ -94,10 +111,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+			await vscode.window.showInformationMessage(messages.graphEngine.dfaRunStopped);
+		} catch (e) {
+			const errMsg = e instanceof Error ? e.message : e as string;
+			outputChannel.appendLine('Failed killing DFA process.');
+			outputChannel.appendLine(errMsg);
+		}
+	} else {
+		await vscode.window.showInformationMessage(messages.graphEngine.noDfaRun);	
+	}
+	void context.workspaceState.update(Constants.GLOBAL_DFA_PROCESS, undefined);
```
jfeingold35, line 128: Question: Is there a reason you're `void`ing this promise instead of `await`ing it or just returning it?

src/extension.ts
```diff
@@ -252,10 +285,23 @@ export async function _runAndDisplayDfa(statusBarItem: vscode.StatusBarItem, run
...
 	}
 }
 
+export async function _shouldProceedWithDfaRun(context: vscode.ExtensionContext, channel: vscode.LogOutputChannel): Promise<boolean> {
+	if (context.workspaceState.get(Constants.GLOBAL_DFA_PROCESS)) {
+		await vscode.window.showInformationMessage(messages.graphEngine.stopDfaRunConfirmationText, messages.graphEngine.stopDfaRunConfirmationYes, messages.graphEngine.stopDfaRunConfirmationNo)
+			.then(async answer => {
+				if (answer === messages.graphEngine.stopDfaRunConfirmationYes) {
+					await _stopExistingDfaRun(context, channel);
+					return true;
```
jfeingold35, line 297: Why does this callback return `true` here? The return value isn't used anywhere as far as I can tell. Also, right now there are cases where it won't return anything, which seems like it's a problem.

src/lib/scanner.ts
```diff
@@ -137,6 +139,30 @@ export class ScanRunner {
         });
     }
 
+    /**
+     * Uses the provided arguments to run a Salesforce Code Analyzer command.
+     * @param args The arguments to be supplied
+     */
+    private async invokeDfaAnalyzer(args: string[], context: vscode.ExtensionContext): Promise<ExecutionResult> {
```
jfeingold35, line 146: I'm prepared to have @stephen-carter-at-sf disagree with me about this, but since this method is almost entirely the same as the `invokeAnalyzer` method, I'd like it if we could either have the `context` be an optional extra parameter for that method, or have them both be wrappers around a third method with the same internal logic. It would cut down on duplicate code.

src/extension.ts
jag-j: I wanted the function to proceed without waiting for the cache update. I also saw this pattern for these cache updates in other repositories.

jfeingold35: Fair enough. Please add a comment in the code explaining that, though.

src/extension.ts
stephen-carter-at-sf: Line 128 has void on it... which is why the compiler is happy. It took me a minute on this one as well... so just for readability/consistency we can remove the void on line 128 and add in : Promise<void> as Josh is suggesting.

src/extension.ts
```diff
@@ -94,10 +111,27 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	return Promise.resolve(context);
 }
 
+export async function _stopExistingDfaRun(context: vscode.ExtensionContext, outputChannel: vscode.LogOutputChannel) {
+	const pid = context.workspaceState.get(Constants.GLOBAL_DFA_PROCESS);
```
stephen-carter-at-sf, line 115: nitpick. Now that it is a workspace setting and not a global setting. Do we want the name of the constant to still be GLOBAL_DFA_PROCESS?

src/lib/scanner.ts
stephen-carter-at-sf: I could go either way with this small amount of code. But if we wanted to share code, then one option is to  have the invokeAnalyzer take a function handle that processes the pid (where its default is a no-op for the standard runs). Then invokeDfaAnalyzer can call the invokeAnalyzer with a function handle that actually does the workspaceState update with the pid/undefined values.

jag-j: I would leave this as is mainly because this whole pid based execution would change to a unified way by calling core in a matter of months.

src/extension.ts
jag-j: Will take care of this.

src/extension.ts
jag-j: Agreed, this should be here.

src/extension.ts
jag-j: I will add Promise<void>. I still want to keep line 128 with void in it. Seems like that's the preference for updating caches (by not blocking the original function by using await).

stephen-carter-at-sf (approved):

PR #71, opened 2024-05-10
FIX (Extension) @W-14001840@: Add tests to increase code coverage and change coverage thresholds

jag-j:


jfeingold35 (approved):

jfeingold35 (approved):

src/test/suite/core-extension-service.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    suite('isAboveMinimumRequiredVersion()', () => {
+        test('returns true when actual version is higher than min version', () => {
+            expect(CoreExtensionService.isAboveMinimumRequiredVersion('1.0.0', '1.0.2')).to.equal(true);
+        });
+
+        test('returns false when actual is less than min version', () => {
+            expect(CoreExtensionService.isAboveMinimumRequiredVersion('1.0.0', '0.0.9')).to.equal(false);
+        });
+
+        test('returns true when actual and min versions are equal', () => {
```
jfeingold35, line 23: The name for this test is a little confusing at first, because it says "equal" but the values being passed in aren't equal. I was able to surmise that it's because "1.0.1" is the first version that's above "1.0.0", but that's not immediately clear. Could you add a comment?

PR #72, opened 2024-05-16
NEW (Extension) @W-14001840@: Change text based on doc review + fix race condition issues with handling DFA process cache

jag-j:


src/extension.ts
```diff
@@ -288,15 +293,10 @@ export async function _runAndDisplayDfa(context:vscode.ExtensionContext ,runInfo
...
+export async function _shouldProceedWithDfaRun(context: vscode.ExtensionContext): Promise<boolean> {
 	if (context.workspaceState.get(Constants.WORKSPACE_DFA_PROCESS)) {
-		await vscode.window.showInformationMessage(messages.graphEngine.stopDfaRunConfirmationText, messages.graphEngine.stopDfaRunConfirmationYes, messages.graphEngine.stopDfaRunConfirmationNo)
-			.then(async answer => {
-				if (answer === messages.graphEngine.stopDfaRunConfirmationYes) {
-					await _stopExistingDfaRun(context, channel);
-				}
-			});
-			return false;
+		await vscode.window.showInformationMessage(messages.graphEngine.existingDfaRunText);
```
jag-j, line 298: I am no longer giving an direct option to cancel the previous run when subsequent runs are triggered. The user has to click on the statusbar and cancel the previous run. During testing I found allowing cancellation from multiple places resulted in unpredictable behavior sometimes. Hence this change.

jfeingold35 (approved):

src/extension.ts
```diff
@@ -116,16 +121,19 @@ export async function _stopExistingDfaRun(context: vscode.ExtensionContext, outp
 	if (pid) {
 		try {
 			process.kill(pid as number, SIGKILL);
+			void context.workspaceState.update(Constants.WORKSPACE_DFA_PROCESS, undefined);
 			await vscode.window.showInformationMessage(messages.graphEngine.dfaRunStopped);
 		} catch (e) {
+			// Exception is thrown by process.kill if between the time the pid exists and kill is executed, the process
+			// ends by itself. Ideally it should clear the cache, but doing this as an abundant of caution.
+			void context.workspaceState.update(Constants.WORKSPACE_DFA_PROCESS, undefined);
```
stephen-carter-at-sf, line 129: Since this is in both branches, does it make sense to put it in a finally block?

jag-j: I want the cache update to happen before the information message is shown and as soon as the process is killed. The reason is to try and avoid any race condition possible. I once hit an issue where the process was all killed and the cache was not cleared and this made starting another DFA run impossible (unless the project is reloaded). I made these small changes of moving the code around to clear the cache and haven't seen any issues since then.

PR #73, opened 2024-05-17
NEW (DevOps) @W-15619625@ Added GHA to create release branch.

jfeingold35:


.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+  schedule:
+    # Cron syntax is "minute[0-59] hour[0-23] date[1-31] month[1-12] day[0-6]". '*' is 'any value,' and multiple values
+    # can be specified with comma-separated lists. All times are UTC.
+    # So this expression means "run at 12 PM UTC, every Friday".
+    - cron: "0 12 * * 5"
+
+
+jobs:
+  # Depending on circumstances, we may want to exit early instead of running the workflow to completion.
+  verify-should-run:
```
jfeingold35, line 24: Direct copy of the corresponding job in `sfdx-scanner`.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+          TODAY_DOW=$(date -u +%u)
+          # This `date` expression returns the last Tuesday of the month, which is our Release Day. %d formats the output
+          # as the day of the month (1-31).
+          NEXT_RELEASE_DATE=$(date -u -v1d -v+1m -v-1d -v-tue +%d)
+          # This `date` expression returns next Tuesday, and `%d` formats the output as the day of the month (1-31).
+          NEXT_TUESDAY_DATE=$(date -u -v+tue +%d)
+          # If the workflow wasn't manually triggered, then it should only be allowed to run to completion on the Friday
+          # before Release Day.
+          [[ $TODAY_DOW != 5 || $NEXT_RELEASE_DATE != $NEXT_TUESDAY_DATE ]] && echo "should_run=false" >> "$GITHUB_OUTPUT" || echo "should_run=true" >> "$GITHUB_OUTPUT"
+  create-release-branch:
```
jfeingold35, line 44: Almost exact copy of the correspondign job in `sfdx-scanner`. Differences called out below.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+              createCommitOnBranch(input: {
+                branch: {
+                  repositoryNameWithOwner: "forcedotcom/sfdx-code-analyzer-vscode",
+                  branchName: $branch
+                },
+                message: {
+                  headline: $message
+                },
+                fileChanges: {
+                  additions: [
```
jfeingold35, line 111: Since there's no RetireJS file, we don't need to submit a change for it.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+      - id: create-branch
+        run: |
+          NEW_VERSION=$(jq -r ".version" package.json)
+          git checkout -b release-$NEW_VERSION
+          # Push the branch immediately without committing anything.
+          git push --set-upstream origin release-$NEW_VERSION
+          # Output the branch name so we can use it in later jobs.
+          echo "branch_name=release-$NEW_VERSION" >> "$GITHUB_OUTPUT"
+      # Update our dependencies
+      - run: yarn upgrade
```
jfeingold35, line 87: I had some concerns about doing this, because in this repo we typically call `yarn install --frozen-lockfile` instead of just `yarn install`. However, upon further inspection, what that does is build without updating the `yarn.lock` file. So it's actually extra important to call `yarn upgrade` here because otherwise we're never going to update the yarn.lock.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+                  id
+                }
+              }
+            }'
+  # Build the scanner tarball so it can be installed locally when we run tests.
+  build-scanner-tarball:
+    name: 'Build scanner tarball'
+    uses: ./.github/workflows/build-scanner-tarball.yml
+    with:
+      # Note: Using `dev` here is technically incorrect. For full completeness's sake, we should probably be
```
jfeingold35, line 133: Drawing attention to this note.

.github/workflows/create-release-branch.yml
```diff
@@ -0,0 +1,146 @@
...
+        options:
+          - major
+          - minor
+          - patch
+        required: true
+  schedule:
+    # Cron syntax is "minute[0-59] hour[0-23] date[1-31] month[1-12] day[0-6]". '*' is 'any value,' and multiple values
+    # can be specified with comma-separated lists. All times are UTC.
+    # So this expression means "run at 12 PM UTC, every Friday".
+    - cron: "0 12 * * 5"
```
jfeingold35, line 19: I debated having this on a delay relative to the scanner, but realized it would only matter if we were building our test tarball from the corresponding scanner release branch, which we currently aren't (see below). If we decide to change that, then I'll change this.

PR #87, opened 2024-05-30
NEW (Extension) @W-14129751@: Allow removal of existing diagnostics

jag-j:


src/extension.ts
```diff
@@ -78,6 +78,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
 		});
 	});
+	const removeDiagnosticsOnActiveFile = vscode.commands.registerCommand(Constants.COMMAND_REMOVE_DIAGNOSTICS_ON_ACTIVE_FILE, (selection: vscode.Uri, multiSelect?: vscode.Uri[]) => {
+		return _clearDiagnosticsForSelectedFiles(multiSelect && multiSelect.length > 0 ? multiSelect : [selection], {
+			commandName: Constants.COMMAND_REMOVE_DIAGNOSTICS_ON_ACTIVE_FILE,
+			diagnosticCollection,
+			outputChannel
+		});
+	});
+	const removeDiagnosticsOnSelectedFile = vscode.commands.registerCommand(Constants.COMMAND_REMOVE_DIAGNOSTRICS_ON_SELECTED_FILE, async (selection: vscode.Uri, multiSelect?: vscode.Uri[]) => {
```
stephen-carter-at-sf, line 88: I see line 81 doesn't have async but this does. Is this intentional?

src/extension.ts
```diff
@@ -78,6 +78,20 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 			outputChannel
 		});
 	});
+	const removeDiagnosticsOnActiveFile = vscode.commands.registerCommand(Constants.COMMAND_REMOVE_DIAGNOSTICS_ON_ACTIVE_FILE, (selection: vscode.Uri, multiSelect?: vscode.Uri[]) => {
```
stephen-carter-at-sf, line 81: I understand that multi select is a think for the selected files from the file tree. But for the active file (which I'm assuming comes up with you right click on a file)... where does multiselect come into play?

I noticed on line 67, our COMMAND_RUN_ON_ACTIVE_FILE doesn't even use the selection or multiSelect arguments. This is making me wonder how that even works because it doesn't even take in the active file as a selection. We just pass in [] there. Hmm.

stephen-carter-at-sf (dismissed):
Approved - but with comments that you might want to consider.

src/extension.ts
jag-j: targeting.getTargets() - line 42 gets the active editor. In order to reach that code, you pass in [] file here. All the file selection logic is encapsulated in that method (that adds some additional logic for Windows). May not be directly useful here, but still that seems to be the pattern I saw in other implementation and followed the same logic.

src/extension.ts
jag-j: No, it wasn't intentional. But I do not know why async is needed here and I see that with other commands as well. I have added it for now, but I think it doesn't really matter.

stephen-carter-at-sf (approved):

PR #91, opened 2024-06-05
NEW (Extension) @W-13991634@ Add class level PMD violations

jag-j:
This PR does the following:
1. Injects PMD's class-level suppression syntax to silence a single rule for that class (i.e., `@SuppressWarnings` syntax).
2. Identifies the specific rule to be suppressed and adds the appropriate syntax (Example: `@SuppressWarnings('PMD.ApexDoc')`).
3. If a class-level suppression tag already exists, it changes the SuppressWarnings accordingly (Example: `@SuppressWarnings('PMD.ApexDoc', 'PMD.UnusedMethod')`), instead of adding a new tag.
4. Adds the tag for Apex and Java appropriately (double-quotes for Java).
5. Identifies the correct position to add this tag by regex parsing the file for 'class', but at the same type ignoring 'class' defined as a comment or as part of a string.

### DIFF
--- a/src/lib/fixer.ts
+++ b/src/lib/fixer.ts
@@ -136,7 +136,7 @@ export class _PmdFixGenerator extends FixGenerator {
         return action;
     }
 
-    private generateClassLevelSuppression(): vscode.CodeAction {
+    public generateClassLevelSuppression(): vscode.CodeAction {
         // Find the end-of-line position of the class declaration where the diagnostic is found.
         const classStartPosition = this.findClassStartPosition(this.diagnostic, this.document);
         // const classEndOfLinePosition = 0;
@@ -153,7 +153,7 @@ export class _PmdFixGenerator extends FixGenerator {
         }
     
         // Extract text from the start to end of the class declaration to search for existing suppressions
-        const classText = this.findLineBeforeClassStartDeclaration(classStartPosition);
+        const classText = this.findLineBeforeClassStartDeclaration(classStartPosition, this.document);
         const suppressionRegex = /@SuppressWarnings\s*\(\s*'([^']*)'\s*\)/;
         const suppressionMatch = classText.match(suppressionRegex);
     
@@ -177,8 +177,8 @@ export class _PmdFixGenerator extends FixGenerator {
     
         action.diagnostics = [this.diagnostic];
         action.command = {
-            command: Constants.COMMAND_RUN_ON_SELECTED,
-            title: 'Re-run diagnostic for this file',
+            command: Constants.COMMAND_REMOVE_DIAGNOSTICS_ON_SELECTED_FILE,
+            title: 'Remove diagnostics for this file',
             arguments: [this.document.uri]
         };
 
@@ -190,7 +190,7 @@ export class _PmdFixGenerator extends FixGenerator {
      * Assumes that the class declaration starts with the keyword "class".
      * @returns The position at the start of the class.
      */
-    private findClassStartPosition(diagnostic: vscode.Diagnostic, document: vscode.TextDocument): vscode.Position {
+    public findClassStartPosition(diagnostic: vscode.Diagnostic, document: vscode.TextDocument): vscode.Position {
         const text = document.getText();
         const diagnosticLine = diagnostic.range.start.line;
     
@@ -221,11 +221,11 @@ export class _PmdFixGenerator extends FixGenerator {
      * Assumes that the class declaration starts with the keyword "class".
      * @returns The text of the line that is one line above the class declaration.
      */
-    private findLineBeforeClassStartDeclaration(classStartPosition: vscode.Position): string {
+    public findLineBeforeClassStartDeclaration(classStartPosition: vscode.Position, document: vscode.TextDocument): string {
         // Ensure that there is a line before the class declaration
         if (classStartPosition.line > 0) {
             const lineBeforeClassPosition = classStartPosition.line - 1;
-            const lineBeforeClass = this.document.lineAt(lineBeforeClassPosition);
+            const lineBeforeClass = document.lineAt(lineBeforeClassPosition);
             return lineBeforeClass.text;
         }
 
--- a/src/lib/fixer.ts
+++ b/src/lib/fixer.ts
@@ -98,6 +98,7 @@ export class _PmdFixGenerator extends FixGenerator {
         const fixes: vscode.CodeAction[] = [];
         if (this.documentSupportsLineLevelSuppression()) {
             fixes.push(this.generateLineLevelSuppression());
+            fixes.push(this.generateClassLevelSuppression());
         }
         return fixes;
     }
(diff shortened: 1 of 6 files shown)


src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+            }
+        } else {
+            // If @SuppressWarnings does not exist, insert a new one
+            const newSuppression = this.generateNewSuppressionTag(suppressionRule, this.document.languageId);
+            action.edit.insert(this.document.uri, classStartPosition, newSuppression);
+        }
+    
+        action.diagnostics = [this.diagnostic];
+        action.command = {
+            command: Constants.COMMAND_REMOVE_DIAGNOSTICS_ON_SELECTED_FILE,
```
jag-j, line 185: FYI, I clear the diagnostics for the whole file and this is not ideal when a file has multiple classes. Keeping track of the diagnostics that would be affected by the rule added on the specific class seemed too difficult and the AC for the story did not call for it. Since this PR is already doing too many things, I could still do it as part of a separate story.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
             arguments: [this.document.uri, this.diagnostic]
         };
+
         return action;
     }
+
+    public generateClassLevelSuppression(): vscode.CodeAction {
+        // Find the end-of-line position of the class declaration where the diagnostic is found.
+        const classStartPosition = this.findClassStartPosition(this.diagnostic, this.document);
+        // const classEndOfLinePosition = 0;
```
stephen-carter-at-sf, line 148: Please remove commented out code.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+        }
+    
+        // Extract text from the start to end of the class declaration to search for existing suppressions
+        const classText = this.findLineBeforeClassStartDeclaration(classStartPosition, this.document);
+        const suppressionMatch = classText.match(this.suppressionRegex);
+    
+        if (suppressionMatch) {
+            // If @SuppressWarnings exists, check if the rule is already present
+            const existingRules = suppressionMatch[1].split(',').map(rule => rule.trim());
+            if (!existingRules.includes(suppressionRule)) {
```
stephen-carter-at-sf, line 168: Just wondering about case sensitivity of the rules. If they are case insensitive, then we probably should add to the map a .toLowerCase() and then add it to the suppressionRule when checking for existence.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+        }
+        return '';
+    }
+
+    /**
+     * Finds the start position of the class in the document.
+     * Assumes that the class declaration starts with the keyword "class".
+     * @returns The position at the start of the class.
+     */
+    public findClassStartPosition(diagnostic: vscode.Diagnostic, document: vscode.TextDocument): vscode.Position {
```
stephen-carter-at-sf, line 216: I was confused at first by this... I didn't realize until I read the code that you always return column 0. This sounds like it is the position (line, col) exactly where you find the word "class". But in reality you are finding the starting line of the class and returning the zeroth column. I recommend you update your comment to make this clear. Otherwise, I was scratching my head earlier in the code wondering how you could put a newline character immediately before the word "class".

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+            // Skip block comment in a single line
+            if (line.match(this.blockCommentEndPattern) && line.match(this.blockCommentStartPattern)) {
+                continue;
+            }
+
+            // Check if this line is the start of a block comment
+            if (line.match(this.blockCommentEndPattern)) {
+                inBlockComment = true;
+                continue;
+            }
```
stephen-carter-at-sf, line 252: I'm a bit concerned how strings are going to prevent these regular expressions from working as you desire them to. For example, if I have a string abc = 'some */ random string'; then it will think you just entered into a block comment (specifically because you are going backward). If you were going forward, the blockCommentStartPattern is more robust.

So I propose you first walk forward through the file and keep track of what lines are comments (since deteching comments blocks are easier going forward). Then afterwards walk backward from the diagnostic line and skip those commented lines. 

Alternatively maybe your isWithinQuotes will help here.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+    /**
+     * Helper function to check if match is within quotes
+     * @param line 
+     * @param matchIndex 
+     * @returns 
+     */
+    public isWithinQuotes(line: string, matchIndex: number): boolean {
+        const beforeMatch = line.slice(0, matchIndex);
+        const singleQuotesBefore = (beforeMatch.match(/'/g) || []).length;
+        const doubleQuotesBefore = (beforeMatch.match(/"/g) || []).length;
```
stephen-carter-at-sf, line 294: Does apex allow double quotes? I thought strings must be single quoted? Not sure if you need to worry about double quotes.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+
+        // Check if the number of quotes before the match is odd (inside quotes)
+        if (singleQuotesBefore % 2 !== 0 || doubleQuotesBefore % 2 !== 0) return true;
+
+        return false;
```
stephen-carter-at-sf, line 299: lower golf score:
return <conditional_statement>

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+        // Find the end-of-line position of the class declaration where the diagnostic is found.
+        const classStartPosition = this.findClassStartPosition(this.diagnostic, this.document);
+        // const classEndOfLinePosition = 0;
+
+        const action = new vscode.CodeAction(messages.fixer.supressOnClass, vscode.CodeActionKind.QuickFix);
+        action.edit = new vscode.WorkspaceEdit();
+    
+        // Determine the appropriate suppression rule based on the type of diagnostic.code
+        let suppressionRule: string;
+        if (typeof this.diagnostic.code === 'object' && this.diagnostic.code !== null && 'value' in this.diagnostic.code) {
```
jfeingold35, line 155: Do you need both the `=== 'object'` and `!== null` check? Seems like the first would be sufficient. Also, in JS/TS it's typical to do `== null` instead of `=== null`, because `undefined == null` and usually you want to catch both null and undefined.

src/lib/fixer.ts
jfeingold35: Follow-up question: Is it an error case if the rule is already present, or is it valid and nothing should happen?

src/lib/fixer.ts
jfeingold35: Clarification: Does "multiple classes" mean multiple outer classes, or an outer class with inner classes? I know Apex can't have multiple outer classes, and I'm pretty sure that the annotation will only apply to outer classes, though I haven't checked that.

src/lib/fixer.ts
jfeingold35: I agree with Stephen. I'd add tests for this kind of corner case if you can.

src/lib/fixer.ts
jfeingold35: Strings have to be single-quoted. I genuinely don't remember if you're allowed to use double-quotes as string boundaries.
BUT you might want to check whether the string containing escaped single-quotes throws off the algorithm. E.g., `'this is a string. it\'s got a single-quote inside it'`.

src/lib/fixer.ts
```diff
@@ -131,6 +138,164 @@ export class _PmdFixGenerator extends FixGenerator {
...
+     * @param matchIndex 
+     * @returns 
+     */
+    public isWithinQuotes(line: string, matchIndex: number): boolean {
+        const beforeMatch = line.slice(0, matchIndex);
+        const singleQuotesBefore = (beforeMatch.match(/'/g) || []).length;
+        const doubleQuotesBefore = (beforeMatch.match(/"/g) || []).length;
+
+        // Check if the number of quotes before the match is odd (inside quotes)
+        if (singleQuotesBefore % 2 !== 0 || doubleQuotesBefore % 2 !== 0) return true;
```
jfeingold35, line 297: Pet peeve: I'm broadly opposed to single-line/braceless `if`s.

src/lib/fixer.ts
jag-j: Didn't realize apex can't have multiple outer classes. I have changed the code and now this applies only to the outer class.

src/lib/fixer.ts
jag-j: I have added case insensitivity in the regex and added tests. It is an error case when a rule is already present and violations are present. One scenario I can think of is when user ran the scanner without the rule and manually added the rule and then tried to run the suppress at the class level. I'm not handling that part.

src/lib/fixer.ts
jag-j: Took care of this.

src/lib/fixer.ts
jag-j: Done.

src/lib/fixer.ts
jag-j: Fixed.

src/lib/fixer.ts
stephen-carter-at-sf: Adding case insensitivity in the regex doesn't help with this line does it existingRules.includes(suppressionRule)) ?

I'd recommend something like
            const existingRules = suppressionMatch[1].split(',').map(rule => rule.trim().toLowerCase());
            if (!existingRules.includes(suppressionRule.toLowerCase())) {

stephen-carter-at-sf (approved):

src/lib/fixer.ts
jag-j: As discussed offline, this has an unintended side effect. It changes all existing rules to lower case since we concatenate the existing rules (which were made to be lower case for comparison) to new rules. So, we decided to leave this as is.

PR #92, opened 2024-06-05
NEW (Extension) @W-13992309@ Remove duplicate suppress violations prompts on the same line

jag-j:
This PR does the following:
1. Shows only one "suppress warning" per line on the editor and in the "Problems" tab per line even if there are multiple violations were found on a line. Previously, multiple "suppress warning"s would show up and they all would add the same "// NOPMD" tag.
2. Additionally, this PR also clears all the diagnostics on the line automatically. We do this by changing the removeSingleDiagnostics command to a more generic removeDiagnosticsInRange that clears all diagnostics in a given range.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -92,8 +92,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 			outputChannel
 		});
 	});
-	const removeSingleDiagnostic = vscode.commands.registerCommand(Constants.COMMAND_REMOVE_SINGLE_DIAGNOSTIC, (uri: vscode.Uri, diagnostic: vscode.Diagnostic) => {
-		_removeSingleDiagnostic(uri, diagnostic, diagnosticCollection);
+	const removeDiagnosticsInRange = vscode.commands.registerCommand(Constants.COMMAND_DIAGNOSTICS_IN_RANGE, (uri: vscode.Uri, range: vscode.Range) => {
+		_removeDiagnosticsInRange(uri, range, diagnosticCollection);
 	});
 	outputChannel.appendLine(`Registered command as part of sfdx-code-analyzer-vscode activation.`);
 	registerScanOnSave(outputChannel);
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -92,8 +92,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 			outputChannel
 		});
 	});
-	const removeSingleDiagnostic = vscode.commands.registerCommand(Constants.COMMAND_REMOVE_SINGLE_DIAGNOSTIC, (uri: vscode.Uri, diagnostic: vscode.Diagnostic) => {
-		_removeSingleDiagnostic(uri, diagnostic, diagnosticCollection);
+	const removeDiagnosticsInRange = vscode.commands.registerCommand(Constants.COMMAND_DIAGNOSTICS_IN_RANGE, (uri: vscode.Uri, range: vscode.Range) => {
```
jag-j, line 95: For a better user experience, we should remove all diagnostics on a given line (or even a set of lines). That's because "// NOPMD" is generic and not specific to a rule. So, a line could have both apexdoc and unused var error and both should go away with the "// NOPMD" tag. This wasn't specifically called out in the acceptance criteria or anywhere else, but I think this is the right thing to do.

src/lib/fixer.ts
```diff
@@ -94,10 +96,14 @@ export class _PmdFixGenerator extends FixGenerator {
      * Generate an array of fixes, if possible.
      * @returns
      */
-    public generateFixes(): vscode.CodeAction[] {
+    public generateFixes(processedLines: Set<number>): vscode.CodeAction[] {
         const fixes: vscode.CodeAction[] = [];
         if (this.documentSupportsLineLevelSuppression()) {
-            fixes.push(this.generateLineLevelSuppression());
+            const lineNumber = this.diagnostic.range.start.line;
```
jag-j, line 102: This is the main change. Even if we have multiple violations on a single line, we show only one "suppress violation" and that would add the "// NOPMD" tag.

stephen-carter-at-sf (approved):

PR #100, opened 2024-07-03
NEW (Extension) @W-15639759@ Introduce a new command to run graph engine at the project level

jag-j:


package.json
```diff
@@ -162,6 +166,10 @@
                     "command": "sfca.runDfaOnSelectedMethod",
                     "when": "false"
                 },
+                {
+                    "command": "sfca.runDfa",
+                    "when": "false"
```
jag-j, line 171: I'm setting this as false so the new command will not show up until Delta run is fully complete. To test this locally, you'll have to change this to true.

src/extension.ts
```diff
@@ -120,19 +120,89 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+		});
+		customCancellationToken = new vscode.CancellationTokenSource();
+		customCancellationToken.token.onCancellationRequested(async () => {
+			customCancellationToken?.dispose();
+			customCancellationToken = null;
+			await vscode.window.showInformationMessage(messages.graphEngine.noViolationsFound);
+			return;
+		});
+
+		const projectDir: string[] = vscode.workspace.workspaceFolders?.map(folder => folder.uri.path);
```
jag-j, line 186: workspaceRoot is now deprecated and this seems to be the way to get the project directory's root. I couldn't get much info on this from the docs, but I have tested this and it works as expected.

src/extension.ts
```diff
@@ -120,19 +120,89 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+		await _runDfa(context, outputChannel);
+	});
+	context.subscriptions.push(runOnActiveFile, runOnSelected, runDfaOnSelectedMethodCmd, runDfaOnWorkspaceCmd, removeDiagnosticsOnActiveFile, removeDiagnosticsOnSelectedFile, removeDiagnosticsInRange);
 	TelemetryService.sendExtensionActivationEvent(extensionHrStart);
 	outputChannel.appendLine(`Extension sfdx-code-analyzer-vscode activated.`);
 	return Promise.resolve(context);
 }
 
+async function _runDfa(context: vscode.ExtensionContext, outputChannel: vscode.LogOutputChannel) {
+	const choice = await vscode.window.showQuickPick(
```
stephen-carter-at-sf, line 146: Does this prompt show up every time - even when you haven't ran before? Ideally it should only show up when we have ran already with violations.

src/extension.ts
```diff
@@ -120,19 +120,89 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+		customCancellationToken.token.onCancellationRequested(async () => {
+			customCancellationToken?.dispose();
+			customCancellationToken = null;
+			await vscode.window.showInformationMessage(messages.graphEngine.noViolationsFound);
+			return;
+		});
+
+		const projectDir: string[] = vscode.workspace.workspaceFolders?.map(folder => folder.uri.path);
+		if (projectDir.length === 0) {
+			void vscode.window.showWarningMessage('***No project directory could be identified. Not proceeding with DFA run.***');
```
stephen-carter-at-sf, line 188: When is this possible? Is it when someone opens up vscode for a file and then closes the file or something?

src/extension.ts
stephen-carter-at-sf: Should we move this logic inside of targeting.getProjectDir() and make currentFile an optional input argument?

jag-j: That's good idea. Done.

src/extension.ts
jag-j: As discussed in the standup, took care of this to show only when cache exists.

src/extension.ts
jag-j: This happens when you haven't opened up any project and still try to run SFCA commands. This is something we should fix at a later point. For example, some of the other extensions make sure a sfdx project is loaded before showing their commands.

jfeingold35 (approved):

PR #103, opened 2024-07-12
NEW (Extension) @W-16095938@ Remove files and directories not required by vsix

jag-j:


.vscodeignore
```diff
@@ -8,3 +8,8 @@ vsc-extension-quickstart.md
 **/.eslintrc.json
 **/*.map
 **/*.ts
+CODEOWNERS
+code-fixtires/**
```
jfeingold35, line 12: Typo, should be `code-fixtures`.

.vscodeignore
```diff
@@ -8,3 +8,8 @@ vsc-extension-quickstart.md
...
+coverage/**
+github-actions/**
```
jfeingold35, line 14: If we're putting `github-actions` here, do we also want to put `.git2gus` and `.github`?

jfeingold35 (changes requested):

.vscodeignore
jag-j: Fixed

.vscodeignore
jag-j: Yes, they can go away as well.

jfeingold35 (approved):

PR #104, opened 2024-07-17
NEW (Extension) @W-16096144@ Show Apex Guru command on explorer when org has Apex Guru perm enabled

jag-j:


src/extension.ts
```diff
@@ -142,6 +149,22 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
...
+		const response:ApexGuruAuthResponse = await connection.request({
+			method: 'GET',
+			url: Constants.APEX_GURU_AUTH_ENDPOINT,
+			body: ''
+		});
+		return response.status == 'Success';
+	} catch(e) {
+		// This could throw an error for a variety of reasons. The API endpoint has not been deployed to the instance, org has no perms, timeouts etc,.
+		// In all of these scenarios, we return false.
+		return false;
```
stephen-carter-at-sf, line 164: This would at the very least be a nice thing to log at some point ... so that if we can't figure out what went wrong - we have somewhere to look.

src/extension.ts
```diff
@@ -142,6 +149,22 @@ export async function activate(context: vscode.ExtensionContext): Promise<vscode
 	return Promise.resolve(context);
 }
 
+export async function _isApexGuruEnabledInOrg(): Promise<boolean> {
+	try {
+		const connection = await CoreExtensionService.getConnection();
+		const response:ApexGuruAuthResponse = await connection.request({
+			method: 'GET',
+			url: Constants.APEX_GURU_AUTH_ENDPOINT,
```
stephen-carter-at-sf, line 157: Oh this is interesting... the CoreExtensionService's connection only requires relative url paths? That's nice.

src/lib/core-extension-service.ts
```diff
@@ -75,6 +79,13 @@ export class CoreExtensionService {
 		}
 	}
 
+	private static initializeWorkspaceContext(workspaceContext: WorkspaceContext | undefined) {
+		if (!workspaceContext) {
+			throw new Error('***workspace context not found***');
+		}
+		CoreExtensionService.workspaceContext = workspaceContext.getInstance(false);
+	}
```
stephen-carter-at-sf, line 87: This will always error if someone installs code analyzer without the extension pack it seems (because the coreExtensionApi can't be pulled in since it is pulled on line 55 using the CORE_EXTENSION_ID.

At least with the telemetry service, we don't error out of it doesn't exist. Shouldn't we do something similar here?

src/lib/core-extension-service.ts
```diff
@@ -139,5 +163,27 @@ interface CoreTelemetryService {
...
+	alias(): string | undefined;
+}
+
+interface Connection {
+	instanceUrl: string;
+	getApiVersion(): string;
+	getUsername(): string | undefined;
+	getAuthInfoFields(): AuthFields;
+	request<T>(options: { method: string; url: string; body: string; headers?: Record<string, string> }): Promise<T>;
+}
```
stephen-carter-at-sf, line 187: I understand that we need to define these because we are pulling the CoreServiceApi from line 35. But I wonder... is it frowned upon to have a dependency on that module so we can pull in its interfaces instead of having to reimplement them here.

This pattern just seems like a flaw in the design of how we are getting access to the CoreExtensionApi's services.

Pulling in all those AuthFields for example... it just feels like at some point we are going to get out of sync with the offical types and interfaces that live in that module.

src/extension.ts
jag-j: Yes :)

src/extension.ts
jag-j: Yes, I really wanted to add a log here as well. But since this was the first step in the activation (before the output channels are defined), I decided not to. May be I should move this to a later point and enable some logging.

src/lib/core-extension-service.ts
jag-j: ApexGuru integration will not work if someone does have the core extension. That's because of all the auth dependencies. So, unlike the telemetry service, we do want to error out. We need better messaging around this when we release this. I hope us being part of the extension pack is a good thing, so not bad overall. Like some of the extension, we could put a note that instead of installing the separate extension, encourage users to install the pack.

src/lib/core-extension-service.ts
jag-j: Agree with your concerns. I didn't find the types exported from the core and e4d followed this pattern. This is already a code smell because now we have these types and interfaces defined at two places. AFAIK, types cannot be exported from one extension, just APIs. So the ideal solution would be to bundle these all into a NPM package that both the core and all the consumer extensions depend on. I will initiate a talk with the extension folks on this. We will have to live with this until then, unfortunately.

src/extension.ts
jag-j: Just to add more details - this is what we get by leveraging sfdx core instead of jsforce or axios. It automatically picks up the base url from the sfdx-project.json file.

src/extension.ts
jag-j: Took care of this, along with some refactoring to make logging a field instead of being passed around.

src/lib/core-extension-service.ts
stephen-carter-at-sf: My point is... we shouldn't require it. Right now.. the plugin (once the boolean is flipped to true) is completely unusable because of this error. initializeWorkspaceContext is called always and if they don't have the extension pack, then an error is thrown it seems.

Instead, why not just disable the apexGuru functionality if this can't be found (just like how telemetry services are disabled if it can't be found).

src/lib/core-extension-service.ts
stephen-carter-at-sf: Yes we should definitely talk with them.

Ideally... they wouldn't require their extension to be installed... but instead allow us to mark a module as a dependency to get these services.

src/lib/core-extension-service.ts
```diff
@@ -85,6 +99,19 @@ export class CoreExtensionService {
 		}
 		throw new Error(messages.error.coreExtensionServiceUninitialized);
 	}
+
+	static async getWorkspaceOrgId(): Promise<string | undefined> {
```
stephen-carter-at-sf, line 103: Where is this being used?

src/lib/core-extension-service.ts
```diff
@@ -65,16 +70,25 @@ export class CoreExtensionService {
 	 * @param telemetryService
 	 * @param context
 	 */
-	private static async initializeTelemetryService(telemetryService: CoreTelemetryService | undefined, context: vscode.ExtensionContext): Promise<void> {
+	private static async initializeTelemetryService(telemetryService: CoreTelemetryService | undefined, context: vscode.ExtensionContext, outputChannel: vscode.LogOutputChannel): Promise<void> {
 		if (!telemetryService) {
-			console.log(`Telemetry service not present in core dependency API. Using null instead.`);
+			outputChannel.warn(`Telemetry service not present in core dependency API. Using null instead.`);
+			outputChannel.show();
```
stephen-carter-at-sf, line 76: do we need to warn about the telemetry service instead of logging. Does that outputChannel show the warning to the user on startup? If so, do we care if they know whether or not telemetry is available?
Telemetry is more of a hidden thin in my mind, whereas the apexguru dependencies aren’t. Thoughts?

stephen-carter-at-sf (approved):

PR #115, opened 2024-08-06
NEW (Extension) @W-16371431@ Poll ApexGuru API and show the response as diagnostic

jag-j:


src/test/suite/index.ts
```diff
@@ -80,9 +80,9 @@ export async function run(): Promise<void> {
 	// This is why we manually set it to 0 earlier.
 	await nyc.checkCoverage({
 		branches: 70,
-		lines: 80,
+		lines: 70,
```
jag-j, line 83: This is an interim fix. As I add the quick-fix code and have some end-to-end tests, this will come up again.

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,66 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
     }
 }
 
-export async function initiateApexGuruRequest(selection: vscode.Uri, outputChannel: vscode.LogOutputChannel): Promise<string> {
+export async function pollAndGetApexGuruResponse(connection: Connection, requestId: string) {
+	let queryResponse: ApexGuruQueryResponse;
+	let isPolling = true;
+	const retryInterval = 1000;
+
+	while (isPolling) {
```
jag-j, line 76: I will talk to ApexGuru team and have a maxretry added here based on inputs from them.

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,67 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
+		diagnosticCollection,
+		commandName,
+		outputChannel
+	} = runInfo;
+	const startTime = Date.now();
 	try {
-        const requestId = await initiateApexGuruRequest(selection, outputChannel);
-		// TODO: Logging the request Id for easy QA. Future stories will use this requestId to poll and retrieve the Apex Guru report.
-        outputChannel.appendLine('***Apex Guru request Id:***' + requestId);
+		return await vscode.window.withProgress({
```
stephen-carter-at-sf, line 44: What is "return" for?

I'm assuming this method has a return signature of Promise<void>. So this return just adds confusion to the code.

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,67 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
+		});
     } catch (e) {
         const errMsg = e instanceof Error ? e.message : e as string;
         outputChannel.appendLine('***Apex Guru initiate request failed***');
         outputChannel.appendLine(errMsg);
     }
 }
 
-export async function initiateApexGuruRequest(selection: vscode.Uri, outputChannel: vscode.LogOutputChannel): Promise<string> {
+export async function pollAndGetApexGuruResponse(connection: Connection, requestId: string) {
```
stephen-carter-at-sf, line 71: nitpick, do you mind adding in the return types to the signature? I'm guessing this returns ApexGuruQueryResponse?

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,67 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
+	let queryResponse: ApexGuruQueryResponse;
+	const retryInterval = Constants.APEX_GURU_RETRY_INTERVAL;
+	let retryCount = Constants.APEX_GURU_RETRY_COUNT;
```
stephen-carter-at-sf, line 74: Instead of having a retryCount and a retryInterval..

Can we just have a pollInterval and a timeOut.

Basically, the way javascript works and the way the request works... the pollInterval is just the minimum wait time - not the exact wait time. So we could end up waiting way longer than we should if you base the final timeout on the retry count.

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,67 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
+
+	while (retryCount > 0) {
+		queryResponse = await connection.request({
```
stephen-carter-at-sf, line 77: Does the request method allow us to send in a connection timeout in case the server freezes? Ideally the individual request timeout should always be close to totalTimeout - currentDuration

src/apexguru/apex-guru-service.ts
jag-j: Yes, this is unnecessary. Will remove.

src/apexguru/apex-guru-service.ts
jag-j: I don't think so. I think this could be one of the requirements we could ask from the Apex Guru team.

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,73 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
+
+			if (queryResponse.status == 'success') {
+				return queryResponse;
+			} else {
+				// Add a delay between requests
+				await new Promise(resolve => setTimeout(resolve, retryIntervalInMillis));
+			}
+		} catch (error) {
+            await new Promise(resolve => setTimeout(resolve, retryIntervalInMillis));
+        }
```
stephen-carter-at-sf, line 90: I wouldn't want to lose this error. I don't think we need a catch statement here do we? Do we care to do retries on error? I didn't think we cared. But if we do, then let's not lose this error. Store it as the last error found... so that if we reach line 95 we can throw the actual error.

package.json
```diff
@@ -105,7 +105,7 @@
             },
             {
                 "command": "sfca.runApexGuruAnalysisOnSelectedFile",
-                "title": "***SFDX: Run Apex Guru Analysis***"
+                "title": "***SFDX: Scan for Performance Issues with ApexGuru***"
```
jshackell-sfdc, line 108:
```suggestion
                "title": "SFDX: Scan for Performance Issues with ApexGuru"
```

stephen-carter-at-sf (approved):

src/apexguru/apex-guru-service.ts
```diff
@@ -29,22 +33,70 @@ export async function isApexGuruEnabledInOrg(outputChannel: vscode.LogOutputChan
...
 }
 
-export async function initiateApexGuruRequest(selection: vscode.Uri, outputChannel: vscode.LogOutputChannel): Promise<string> {
+export async function pollAndGetApexGuruResponse(connection: Connection, requestId: string, maxWaitTimeInSeconds: number, retryIntervalInMillis: number): Promise<ApexGuruQueryResponse> {
+	let queryResponse: ApexGuruQueryResponse;
+	let lastErrorMessage = '';
+	const startTime = Date.now();
+	while ((Date.now() - startTime) < maxWaitTimeInSeconds * 1000) {
+		try {
+			queryResponse = await connection.request({
```
stephen-carter-at-sf, line 77: We should not only follow up with the folks who own this request method regarding the response code... but we should also ask if we can please send in a connection timeout value to ensure that if the server goes down... it isn't going past our remaining timeout time.

stephen-carter-at-sf (approved):

PR #116, opened 2024-08-14
NEW (Extension) @W-16442046@ Apex guru fixer logic - to show and insert apex guru suggestions

jag-j:


src/apexguru/apex-guru-service.ts
```diff
@@ -128,16 +127,19 @@ export function transformStringToRuleResult(fileName: string, jsonString: string
...
 			ruleName: parsed.type,
 			message: parsed.value,
 			severity: 1,
 			category: parsed.type, // Replace with actual category if available
-			line: parseInt(parsed.properties.find((prop: ApexGuruProperty) => prop.name === 'line_number')?.value),
+			line: lineNumber,
 			column: 1,
-			suggestedCode: Buffer.from(encodedClassAfter, 'base64').toString('utf8')
+			currentCode: Buffer.from(encodedCodeBefore, 'base64').toString('utf8'),
+			suggestedCode: Buffer.from(encodedCodeAfter, 'base64').toString('utf8'),
```
stephen-carter-at-sf, line 142: If currentCode and suggestedCode is equal to `` (i.e. empty)… does it show this in the popup box still?
I ask because I would have assumed that to trigger whether it shows up or not would be based on undefined or null instead. Because maybe the after code is indeed empty and we want to show that instead of hide it.

src/lib/diagnostics.ts
```diff
@@ -89,6 +89,23 @@ export class DiagnosticManager {
             target: vscode.Uri.parse(violation.url),
             value: violation.ruleName
         } : violation.ruleName;
+        if (engine === 'apexguru') {
+            const apexGuruViolation = violation as ApexGuruViolation;
+        
+            if (apexGuruViolation.suggestedCode?.trim()) {
```
stephen-carter-at-sf, line 95: If the suggestedCode is empty... maybe that is because we are suggesting to replace it. Can we instead just explicitly check if the suggestedCode is undefined as a way of determining this.

src/apexguru/apex-guru-service.ts
jag-j: As explained in slack, this doesn't show up.

src/lib/diagnostics.ts
jag-j: Ah! this is a super-edge case. But valid point. Let me take care of this.

jag-j: Done!

src/apexguru/apex-guru-service.ts
stephen-carter-at-sf: Maybe not today. But it could in the future.

Instead of 
```
(prop: ApexGuruProperty) => prop.name === 'code_after')?.value ?? ''
```

Why not just send in the value always even if it is undefined. 

Then your line
```
if (apexGuruViolation.suggestedCode) {
```
code could just switch to
```
if (apexGuruViolation.suggestedCode !== undefined) {
```
so that it allows for empty strings.

PR #118, opened 2024-08-16
NEW (Extension) @W-16442046@ do not throw error when source is undefined

gbockus-sf:
Kept seeing an error in VSCode randomly associated with CA extension.  Figured I'd take a quick peek to see if I could figure out why. 

Root cause here: https://github.com/forcedotcom/sfdx-code-analyzer-vscode/blob/4ed5252d599d00ba8b7ddb103f9ede07550f794a/src/lib/fixer.ts#L27

I can't really explain why TS isn't complaining about the string | undefined mismatch with the isSource method, but this should prevent the error. 

I didn't see any tests covering this area, but lemme know if I missed it.

### DIFF
--- a/src/lib/messages.ts
+++ b/src/lib/messages.ts
@@ -39,7 +39,13 @@ export const messages = {
         messageGenerator: (severity: number, message: string) => `Sev${severity}: ${message}`,
         source: {
             generator: (engine: string) => `${engine} via Code Analyzer`,
-            isSource: (source: string) => source.endsWith(' via Code Analyzer'),
+            isSource: (source: string) => {
+                if (!source) {
+                    return false;
+                }
+                
+                return source.endsWith(' via Code Analyzer');
+            },
             extractEngine: (source: string) => source.split(' ')[0]
         }
     },


jag-j (approved):

src/lib/messages.ts
```diff
@@ -46,7 +46,13 @@ export const messages = {
...
         source: {
             generator: (engine: string) => `${engine} via Code Analyzer`,
-            isSource: (source: string) => source.endsWith(' via Code Analyzer'),
+            isSource: (source: string) => {
+                if (!source) {
+                    return false;
+                }
+                
+                return source.endsWith(' via Code Analyzer');
+            },
```
stephen-carter-at-sf, line 55: Lower golf score:

isSource: (source: string) => source && source.endsWith(' via Code Analyzer'),

stephen-carter-at-sf (changes requested):
Lower golf

src/lib/messages.ts
jag-j: @gbockus-sf - looks like you need code signing set up. I'll take care of this if that's fine with you. Apologies for the delay - this has been on my todo ever since you reported. I should have gotten to this quicker!

jag-j: @gbockus-sf : Here is the [PR](https://github.com/forcedotcom/sfdx-code-analyzer-vscode/pull/119) with similar fix to another function + tests. @stephen-carter-at-sf: Both your suggestions return undefined instead of false when source is falsy in my tests. This was a surprise to me. I have left the code as is for now. I'll go ahead and close this PR without merging.

PR #119, opened 2024-08-19
NEW (Extension) @W-16442046@ Do not throw error when source is undefined

jag-j:


src/lib/messages.ts
```diff
@@ -46,8 +46,13 @@ export const messages = {
         messageGenerator: (severity: number, message: string) => `Sev${severity}: ${message}`,
         source: {
             generator: (engine: string) => `${engine} via Code Analyzer`,
-            isSource: (source: string) => source.endsWith(' via Code Analyzer'),
-            extractEngine: (source: string) => source.split(' ')[0]
+            isSource: (source: string) => {
```
jag-j, line 49: @stephen-carter-at-sf: Both your suggestions return undefined instead of false when source is falsy in my tests. This was a surprise to me. I have left the code as is for now.

stephen-carter-at-sf (approved):

PR #121, opened 2024-08-20
FIX (Extension) @W-16382598@ Expose ApexGuru as a user settings + ApexGuru usability improvements

jag-j:


package.json
```diff
@@ -158,6 +162,11 @@
                 "codeAnalyzer.rules.category": {
                     "type": "string",
                     "description": "The categories of rules to run. Specify multiple values as a comma-separated list. Run 'sf scanner rule list -e' in the terminal for the list of categories associated with a specific engine."
+                },
+                "codeAnalyzer.apexGuru.enabled": {
+                    "type": "boolean",
+                    "default": false,
```
stephen-carter-at-sf, line 168: Oh false by default. For some reason I thought we were going to just rely on the org perm check (making it async if it wasn't already) as the means to determine enablement by default. So I was expecting this to be true by default. So with this users will first need to enable the setting and then have their org (with perm enabled) authed.

src/lib/diagnostics.ts
```diff
@@ -96,11 +96,11 @@ export class DiagnosticManager {
                 diagnostic.relatedInformation = [
                     new vscode.DiagnosticRelatedInformation(
                         new vscode.Location(vscode.Uri.parse(violation.url), range),
-                        `Current Code: ${apexGuruViolation.currentCode}`
+                        `\n// Current Code: \n ${apexGuruViolation.currentCode}`
```
stephen-carter-at-sf, line 99: Did you mean to have a space after the second newline? `\n `? If so I'm assuming that is just for display purposes?

stephen-carter-at-sf (approved):

src/lib/diagnostics.ts
jag-j: My bad! Fixed this.

package.json
jag-j: Yes, I have been going back and forth on this. But given how small the closed pilot is going to be, I don't want to initialize core service for rest of the users just to check the org perm. So, went ahead with this approach.

stephen-carter-at-sf (approved):

PR #130, opened 2024-09-16
NEW (Extension) @W-15640497@ Delta runs on SFGE - Phase 1

jag-j:


package.json
```diff
@@ -36,6 +36,7 @@
         "cross-spawn": "^7.0.3",
         "glob": "^8.0.3",
         "globby": "^11.0.0",
+        "proxyquire": "^2.1.3",
```
jfeingold35, line 39: Do you have a 3PP for this? Also, it doesn't look like this is used in production code, so should it be in `devDependencies` instead of `dependencies`?

src/deltarun/delta-run-service.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as fs from 'fs';
+
+export function getDeltaRunTarget(sfgecachepath:string, savedFilesCache:Set<string>): string[] {
```
jfeingold35, line 9: Conceptually, it feels weird for this to be a utility method that accepts the `Set`, instead of having a `class` that encloses the `Set` and exposes a `getDeltaRunTarget` method.
But broadly speaking, I think this is fine.

src/extension.ts
```diff
@@ -194,10 +209,15 @@ async function _runDfa(context: vscode.ExtensionContext) {
 		);
 
 		// Default to "Yes" if no choice is made
-		const rerunFailedOnly = choice == '***Yes***';
-		if (rerunFailedOnly) {
+		const rerunChangedOnly = choice == '***Yes***';
+		if (rerunChangedOnly) {
 			// Do nothing for now. This will be implemented as part of W-15639759
```
jfeingold35, line 214: Does this comment need to change?

jag-j: Done

package.json
jag-j: My bad! I intended this to a devDependencies as we discussed in the tea time last week. I have moved it now.

jfeingold35 (approved):

jfeingold35 (approved):

PR #142, opened 2024-10-14
NEW (Extension) @W-16812379@ Add production heartbeat test

jag-j:


.github/workflows/production-heartbeat.yml
```diff
@@ -11,28 +11,148 @@ on:
...
+
+      - name: Install Salesforce Code Analyzer Extension on Windows
+        if: runner.os == 'Windows'
+        run: |
+          echo "Installing Code Analyzer Extension..."
+          "/c/Program Files/Microsoft VS Code/bin/code" --install-extension salesforce.sfdx-code-analyzer-vscode
+          echo "Installing Code Analyzer Complete"
+          
+          echo "Waiting for 10 seconds..."
+          sleep 10
```
stephen-carter-at-sf, line 68: Why do we sleep here? Is the install really not complete?

.github/workflows/production-heartbeat.yml
```diff
@@ -11,28 +11,148 @@ on:
...
+          code --install-extension salesforce.sfdx-code-analyzer-vscode
+
+      - name: Verify Extension Installation on macOS
+        if: runner.os == 'macOS'
+        run: |
+          if code --list-extensions | grep -q 'salesforce.sfdx-code-analyzer-vscode'; then
+            echo "Extension installed successfully"
+          else
+            echo "Extension installation failed" && exit 1
+          fi
```
stephen-carter-at-sf, line 111: These steps seem to be the same as ubuntu. Any reason why we didn't reuse this code?

.github/workflows/production-heartbeat.yml
```diff
@@ -11,28 +11,148 @@ on:
 jobs:
   production-heartbeat:
     strategy:
-      # By default, if any job in a matrix fails, all other jobs are immediately cancelled. This makes the jobs run to completion instead.
       fail-fast: false
       matrix:
-        os: [{vm: ubuntu-latest, exe: .sh}, {vm: windows-2019, exe: .cmd}]
+        os: [{vm: ubuntu-latest, exe: .sh}, {vm: windows-2019, exe: .cmd}, {vm: macos-latest, exe: .sh}]
```
stephen-carter-at-sf, line 16: Where is exe field being used? 
I see we are using the vm field... but I don't see the exe field being used anywhere.

.github/workflows/production-heartbeat.yml
```diff
@@ -11,28 +11,148 @@ on:
...
+          RUN_LINK: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+        run: |
+
+          if [[ ${{ env.IS_CRITICAL }} == true ]]; then
+            ALERT_SEV="critical"
+            ALERT_SUMMARY="Production heartbeat script failed on ${{ runner.os }}"
+          else
+            ALERT_SEV="info"
+            ALERT_SUMMARY="Production heartbeat script succeeded with retries on ${{ runner.os }}"
```
stephen-carter-at-sf, line 135: Does this line of code ever git hit given that the conditional on line 118 is for failure() || cancelled()?

For cancelled... do we really want to say things succeeded?

.github/workflows/production-heartbeat.yml
jag-j: We are doing this on Windows since I noticed that sometimes list-extensions doesn't show the extension right away.

.github/workflows/production-heartbeat.yml
jag-j: Fixed.

.github/workflows/production-heartbeat.yml
jag-j: You are right. It doesn't. But it was useful for all the tests I ran on Friday to test the pager. So, I would probably leave it. (sfdx-scanner heartbeat also follows a similar strategy. )

jfeingold35: I'd at least add a comment indicating that the `else` branch is impossible to hit in normal execution, maybe?

jfeingold35 (approved):

PR #146, opened 2024-10-16
FIX (Extension) @W-16993912@ Implement Doc Review suggestions + fix partial runs setting bug

jag-j:


package.json
```diff
@@ -169,10 +169,10 @@
                     "default": false,
                     "description": "Discover critical problems and performance issues in your Apex code with ApexGuru, which analyzes your Apex files for you. This feature is in a closed pilot; contact your account representative to learn more."
                 },
-                "codeAnalyzer.deltaRuns.enabled": {
+                "codeAnalyzer.partialGraphEngineScans.enabled": {
                     "type": "boolean",
                     "default": false,
-                    "description": "***Enable delta run of Salesforce Graph Engine***"
+                    "description": "Partial Salesforce Graph Engine Scans: Enabled"
```
jag-j, line 175: @jshackell-sfdc: As discussed on slack, can we elaborate a bit more about the feature in the description and also call out that this is a beta feature? Thanks!

jag-j: Discussed offline with Juliet and I have updated the text. Resolving!

stephen-carter-at-sf (approved):
Approved from a coding perspective. I'm assuming Juliet has/will review the text.

PR #151, opened 2024-12-05
FIX (Extension) @W-17311922@ Fix pager duty alert subject

jag-j:


.github/workflows/daily-smoke-test.yml
```diff
@@ -38,16 +38,10 @@ jobs:
       - name: Report problems
         shell: bash
         env:
-          IS_CRITICAL: ${{ contains(join(steps.*.outcome), 'failure') || contains(join(steps.*.outcome), 'skipped') }}
           RUN_LINK: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
-          if [[ ${{ env.IS_CRITICAL }} == true ]]; then
```
jag-j, line 44: The issue seems to be the way we are assigning IS_CRITICAL. There are two ways to fix this: either assign it correctly or altogether remove this. I prefer removing this because I think the reason this was added was for testing. But this isn't greatly helpful since the if condition in line 36 will anyways needs to be altered to test the script. Let me know if you think otherwise.

stephen-carter-at-sf: Agreed... but should we remove the || cancelled() from the if statement then on line 36? What does that do?

jfeingold35: IIRC, there are technically ways that a job can end up in `cancelled()` due to something that would more accurately be described as a failure. But I can't remember what they are off the top of my head.

jag-j: Hmmm, let's err in the side of caution and leave cancelled() as is.

stephen-carter-at-sf (approved):

PR #155, opened 2025-01-17
NEW (Extension) @W-17330890@ A4D SFCA Integration - Integrate with UnifiedDiff UI

jag-j:


package.json
```diff
@@ -33,7 +33,9 @@
         "SFCA"
     ],
     "dependencies": {
+        "@salesforce/vscode-service-provider": "^1.3.0-rc.6.2",
         "cross-spawn": "^7.0.3",
+        "einstein-shared": "https://github.com/forcedotcom/salesforcedx-vscode-einstein-shared#semver:0.3.9",
```
jag-j, line 38: This is the shared module repository and since the package cannot be open sourced yet, we depend on it this way.

stephen-carter-at-sf: When installing our vscode extention… will the new dependency
einstein-shared": "https://github.com/forcedotcom/salesforcedx-vscode-einstein-shared#semver:0.3.9",
get pulled during install or is it installed into the vsix file?

That is... can we even ship this line of code or will it break users during install?

jag-j: We can ship with this line of code. This is bundled into the vsix.

stephen-carter-at-sf (approved):

jfeingold35 (approved):

PR #159, opened 2025-01-29
NEW (Extension) @W-17682771@ Interim fix to parse and extract code from llm response

jag-j:


src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -102,12 +102,21 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
         })();
     }
 
-    public removeCodeMarkdowns(codeSnippet: string) {
-        if (codeSnippet.startsWith('```apex')) {
-            codeSnippet = codeSnippet.trimEnd();
-            codeSnippet = codeSnippet.slice(7, codeSnippet.length - 3);
+    public extractCodeFromResponse(response: string) {
+        const startTag = '```apex';
+        const endTag = '``';
```
stephen-carter-at-sf, line 107: This seems wrong... should be triple ticks right?

Besides, why not just use a regex to extract the apex code between these tags. For example:
```
const matches = [...input.matchAll(```apex([\s\S]*?)```/g)];
```
and then just grab the first match

stephen-carter-at-sf (changes requested):

src/modelBasedFixers/apex-pmd-violations-fixer.ts
jag-j: The model is inconsistent with the code block. I sometimes see double backticks as the end of the codeblock. We should bring this up with the Ai team in the next meeting.
I avoid regex when possible - mainly readability/maintainability and in this case, this is going to change based on discussions with regex and want something that is verbose. Do you think any advantage regex brings here ?

stephen-carter-at-sf: extracting strings between tokens is what regex is best for in my opinion.

But if you don't go with regex, then can you please add a comment to line 107 about the issue with the LLM response sometimes not giving triple backticks.

stephen-carter-at-sf (approved):

PR #161, opened 2025-02-05
NEW (Extension) @W-17663361@ Remove the diagnostics for approved/approved-all code fixes

jag-j:


src/extension.ts
```diff
@@ -592,3 +597,6 @@ export function _isValidFileForAnalysis(documentUri: vscode.Uri) {
 	return allowedFileTypes.includes(path.extname(documentUri.path));
 }
 
+export function extractDiagnosticCode(diagnostic: vscode.Diagnostic) {
+	return typeof diagnostic.code === 'object' && 'value' in diagnostic.code ? String(diagnostic.code.value) : '';
```
jfeingold35, line 601: IIRC, using the `String` constructor is generally discouraged. Is there a `.toString()` method you can call, or something similar?

jfeingold35 (approved):

PR #163, opened 2025-02-07
CHANGE (Extension) @W-17683398@ Refactored into a strategy-based framework

jfeingold35:
This PR is the first of two PRs comprising work item W-17683398.
It refactors the existing code into a strategy pattern, and makes other related refactors for improved testability.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,14 +9,16 @@
 import * as vscode from 'vscode';
 import * as targeting from './lib/targeting';
 import {ScanRunner} from './lib/scanner';
-import {SettingsManager} from './lib/settings';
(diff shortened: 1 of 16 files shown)


src/extension.ts
```diff
@@ -597,3 +610,27 @@ export function _isValidFileForAnalysis(documentUri: vscode.Uri) {
 	return allowedFileTypes.includes(path.extname(documentUri.path));
 }
 
+class VSCodeDisplayable implements Displayable {
```
jfeingold35, line 613: This class can abstract away all of the VSCode-specific stuff about UI/UX, so other classes won't need to interact directly with VSCode as much.

PR #171, opened 2025-02-21
NEW (Extension)  @W-17831303@ Fix release branch creation issue

jag-j:


.github/workflows/create-release-branch.yml
```diff
@@ -60,7 +60,9 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: 'lts/*' # Always use Node LTS for building dependencies.
-      - run: yarn
```
jag-j, line 63: Not sure why we use yarn here instead of using --frozen-lockfile. Let me know if this was intentional.

.github/workflows/create-release-branch.yml
```diff
@@ -60,7 +60,9 @@ jobs:
...
+      - name: Configure Git for private repo access
+        run: git config --global --add url."https://${{ secrets.IDEE_GH_TOKEN }}@github.com/".insteadOf "https://github.com/"
```
stephen-carter-at-sf, line 64: For the preceding
```
    uses: actions/checkout@v4
```
step in other files, we have in 
```
        with:
          token: ${{ secrets.IDEE_GH_TOKEN }} # So we can access internal repo to pull shared ui components
```
Why is this needed in the other files?

Do we need to have that here in this file as well?

.github/workflows/create-release-branch.yml
jfeingold35: Using `--frozen-lockfile` throughout probably isn't the worst idea.

.github/workflows/publish.yml
```diff
@@ -82,6 +82,8 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '11'
+      - name: Configure Git for private repo access
+        run: git config --global --add url."https://${{ secrets.IDEE_GH_TOKEN }}@github.com/".insteadOf "https://github.com/"
```
stephen-carter-at-sf, line 86: ditto... about the actions/checkout@v4 stuff

stephen-carter-at-sf (commented):
I wonder if it is worth creating a composite github action that for the checkout, node setup, and yarn install that we can just share all this across all our actions as a single looking action step.

.github/workflows/create-release-branch.yml
stephen-carter-at-sf: As discussed offline... we should instead try to remove all this git config update stuff... in favor of just using the token during checkout. But I believe this will require you to update the dependencies in the package.json file with 
```
-        "einstein-shared": "https://github.com/forcedotcom/salesforcedx-vscode-einstein-shared#semver:0.3.10",
+        "salesforcedx-vscode-einstein-shared": "forcedotcom/salesforcedx-vscode-einstein-shared#semver:0.3.10",
```
and then rerun yarn install so that it updates the yarn.lock file. Then submit those updates and see if it fixes everything for us.

stephen-carter-at-sf (approved):

PR #175, opened 2025-02-24
NEW (Extension)  @W-17831303@ Remove dependency on internal shared repo

jag-j:


stephen-carter-at-sf (approved):

.github/workflows/publish.yml
```diff
@@ -59,7 +59,6 @@ jobs:
       # Before publishing, we want to test the extension against whatever
       # version of the scanner is currently live.
       use-scanner-tarballs: false
-    secrets: inherit
```
jfeingold35, line 62: I think this might actually be necessary, so I recommend leaving it in.

.github/workflows/validate-pr.yml
```diff
@@ -38,10 +38,8 @@ jobs:
       use-scanner-tarballs: true
       v4-tarball-suffix: 'dev-4'
       v5-tarball-suffix: 'dev'
-    secrets: inherit
```
jfeingold35, line 41: I recommend leaving this in. I'm not positive it's necessary here, but I think it might be.

.github/workflows/publish.yml
jag-j: As discussed offline, this was something I added to fetch the internal repo. We can remove this.

PR #180, opened 2025-02-25
NEW (Extension) @W-17865130@ Fix formatting and multiples violations in a single line issue

jag-j:


src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -28,7 +28,7 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
         const filteredDiagnostics = context.diagnostics.filter(
             diagnostic => messages.diagnostics.source
               && messages.diagnostics.source.isSource(diagnostic.source)
-              && diagnostic.range.isEqual(range)
+              && range.contains(diagnostic.range)
```
jag-j, line 31: When multiple diagnostics are on the same line, the range and diagnostic need not align. The range is calculated from the user interaction and we should just care about that the diagnostic for which we are providing the code fix is within the range of the user selection.

src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -195,6 +191,64 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
         return updatedLines.join(lineEnding);
     }
 
+    private formatCode(
```
jag-j, line 194: This is a best effort formatting at this point. I deformat the lines we get as suggestion from LLM and apply format based on the files content and based on curly brackets.

src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -195,6 +191,64 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
...
+            
+            // Adjust for current line
+            if (line.startsWith('}')) {
+                indentLevel--;
+            }
+
+            // Apply document's base indentation plus additional levels
+            const indentation = baseIndentation + '    '.repeat(Math.max(0, indentLevel));
+            return indentation + line;
+        });
```
stephen-carter-at-sf, line 242: You have a loop in a loop that seems unnecessary.

Instead, why not put the indentLevel on the outside and as you walk each line... if you see a { on the current line then increase the indent level and the end of the loop for the next line to consume. And if you see a } on the current line then decrease the indent level at the end of the loop.

Then your algorithm be O(n) instead.
The only edge case is take into account when { and } both appear on the same line... which I believe will be handled automatically as long as you keep the previous 2 conditions separate.

stephen-carter-at-sf (changes requested):

src/test/suite/modelBasedFixers/apex-pmd-violations-fixer.test.ts
```diff
@@ -8,6 +8,7 @@
 import * as vscode from 'vscode';
 import { expect } from 'chai';
 import { ApexPmdViolationsFixer } from '../../../modelBasedFixers/apex-pmd-violations-fixer';
+import Sinon from 'sinon';
```
stephen-carter-at-sf, line 11: Not blocking... but just saying... I really dislike Sinon... and instead favor creating interfaces between dependencies like this.

Feel free to continue with this, but at some point I might go in and put an indirection layer in the mix to help with unit testing. Having our own interface with only the getters that we need with a runtime instance that composes a vscode.TextDocument is preferable so that we instead can just create a test double for our interface.

jfeingold35: Agree. Particularly in a context like VSCode, Sinon will have you stubbing out tons of stuff. the interface/instance paradigm works way better here.

src/modelBasedFixers/apex-pmd-violations-fixer.ts
jag-j: Good catch! Sometimes when you work one usecase at a time, you miss the larger picture. The code is much simpler now.

src/test/suite/modelBasedFixers/apex-pmd-violations-fixer.test.ts
jag-j: +1. I could have gone either ways. But looking at other tests using sinon, I went with this for consistency.

jfeingold35: That's certainly fair. When we have a minute, the tests are something that could benefit from an overhaul.

src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -195,6 +191,55 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
...
+                return '';
+            }
+            if (line.startsWith('}')) {
+                indentLevel = Math.max(0, indentLevel - 1);
+            }
+
+            const indentation = baseIndentation + '    '.repeat(indentLevel);
+
+            // Adjust for next lines, this line needs no changes
+            if (line.includes('{')) {
```
jfeingold35, line 228: It's technically possible to do something like `if (x) {System.debug('asdfasdfasdf');}` all on one line, so I think you want to do `line.endsWith('{'})` to make sure you properly handle that case.

jag-j: Ok, I knew edge cases like that are possible. I took care of } <some code> {, but now this. I'll take care of this.

jfeingold35 (approved):

PR #181, opened 2025-02-27
NEW (Extension)  @W-17865130@ Fix indentation for paranthesis and square brackets

jag-j:


jfeingold35 (approved):

stephen-carter-at-sf (approved):

src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -227,6 +243,10 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
             // Adjust for next lines, this line needs no changes
             if (line.endsWith('{')) {
                 indentLevel++;
+            } else if (line.endsWith('(')) {
+                parenLevel++;
+            } else if (line.endsWith('[')) {
+                braceLevel++;
             }
```
stephen-carter-at-sf, line 250: My hunch is that this is wrong. I believe you probably should be doing an includes... or even more robust would be to walk character by character instead of line by line keeping track.

Otherwise an edge case like:
this.hello(
    arg1,
    arg2)
this.world()

will end up looking like:
this.hello(
    arg1,
    arg2)
    this.world()

stephen-carter-at-sf (changes requested):

src/modelBasedFixers/apex-pmd-violations-fixer.ts
stephen-carter-at-sf: Also if a comment is in the mix:
```
function hello() { // this is a comment
   String a = 3;
}
```

will end up looking like
```
function hello() { // this is a comment
String a = 3;
}
```

stephen-carter-at-sf: And note we'll have issues if we aren't also keeping track of comments... for example:
```
String a = 3; // This is a comment {
String b = 4; // } this is also a comment
```
will end up being:
```
String a = 3; // This is a comment {
    String b = 4; // } this is also a comment
```

stephen-carter-at-sf: I don't know if we are going to get things right. This is why I think my approach might be best...


Just shift everything left by the amount found in the minimum indent. For example, first take:
```
    someFunc() {
       def();
    }
```
and shift it all by 3 (which is the smallest indent on all the lines) so that it first is:
```
someFunc() {
   def();
}
```
then shift it to the right until the first line was indented the same as it was before. For example if it originally was:
```
                someCodeBeforeTheFixCameIn();
                some other code;
```
then we would make the resulting lines have to shift by `                ` and thus we would have:
```
                someFunc() {
                   def();
                }
```
========
Another example: 
if the original code is:
```
        someOriginalCode() // starts with 8 spaces
    } else {
        // no op
    }
```
then if the reply from A4D was:
```
              someFixedCode() // For some odd reason starts with 14 spaces
          } else {
              // no op
          }
```
then we first left indent everything that we got from A4D:
```
    someFixedCode() // After shifting left, there is now 4 spaces
} else {
    // no op
}
```
and then in order to get back to 8 spaces, we see we need to indent everything by 4 spaces right:
```
        someFixedCode() // After shifting left, there is now 4 spaces
    } else {
        // no op
    }
```

----
Otherwise, I think writing a formatter is going to be too difficult because there are going to be too many edge cases to handle with comments, etc.

jag-j: Agreed that model returning the correct indentation and we doing the relative indentation is the best approach. As I said in the slack thread - any of these edge cases are far better than the model returning incorrect indentation and things looking bad all through. I would prefer to leave this as is and we start the conversation with the AI team on formatted response.

stephen-carter-at-sf: OK fair enough.

stephen-carter-at-sf (approved):

PR #183, opened 2025-03-03
CHANGE (misc) @W-17950421@ Update dependencies, switch to npm, and fix lint warnings

stephen-carter-at-sf:


eslint.config.mjs
```diff
@@ -0,0 +1,16 @@
+import eslint from '@eslint/js';
```
stephen-carter-at-sf, line 1: Using new flat config with eslint 9... thus we no longer have a .estlintrc.json file and have this instead.

package.json
```diff
@@ -17,7 +17,8 @@
     "publisher": "salesforce",
     "license": "BSD-3-Clause",
     "engines": {
-        "vscode": "^1.82.0"
+        "vscode": "^1.90.0",
+        "node": ">=20.9.0"
```
stephen-carter-at-sf, line 21: Peter Hale said we can move our minimum support to 1.90.0 for vscode (which also bumps up node). For node,  we can allow for major versions that are greater than 20, but i'm not sure if we will allow for vscode major versions or not because they haven't release version 2, etc. So not sure if it's best practice to have "^" or ">=" here... i'm assuming that this is safer.

src/lib/targeting.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import * as vscode from 'vscode';
-import globby = require('globby');
+import {glob} from 'glob';
```
stephen-carter-at-sf, line 8: I noticed we had both globy and globby originally in this repo. I removed globby and consolidated with just glob. Note globby's latest is now ESM module... so that is why I chose glob over globby to consolidate to - because it plays nicer with commonJS modules.

src/lib/targeting.ts
```diff
@@ -30,9 +30,10 @@ export async function getTargets(selections: vscode.Uri[]): Promise<string[]> {
 				// Globby wants forward-slashes, but Windows uses back-slashes, so we need to convert the
 				// latter into the former.
 				const globbablePath = selection.fsPath.replace(/\\/g, '/');
-                const globOut: string[] = await globby(`${globbablePath}/**/*`);
+                const globOut: string[] = await glob(`${globbablePath}/**/*`, {nodir: true});
```
stephen-carter-at-sf, line 33: glob is different than globby in that it includes the directories as well as the files by default. Thus I had to add in this nodir flag to maintain existing behavior.

src/test/suite/apexguru/apex-guru-service.test.ts
```diff
@@ -288,7 +288,7 @@ suite('Apex Guru Test Suite', () => {
 
             // ===== ASSERTIONS =====
             expect(response).to.deep.equal(queryResponse);
-            expect(connectionStub.request.calledOnce).to.be.true;
+            expect(connectionStub.request.calledOnce).to.equal(true);
```
stephen-carter-at-sf, line 291: This is an example of where chai (even though I didn't update it) is now being flagged by the modern eslint rules. I don't think we were suppose to be doing "to.be.true" but it allows us to. I switch many of these where I could.

src/test/suite/extension.test.ts
```diff
@@ -1,14 +1,14 @@
+/* eslint-disable @typescript-eslint/no-unused-expressions */ // TODO: Need to update these old tests... many of the chair assertions are not being used correctly causing eslint errors.
```
stephen-carter-at-sf, line 1: Just too many chai eslint warnings to deal with... and since this test will be reworked in the near future - just suppressing them for now.

eslint.config.mjs
jfeingold35: It looks like we lose the type checking rules in this transition. Should we not be using `tseslint.configs.recommendedTypeChecked`?

package.json
```diff
@@ -26,54 +27,52 @@
...
+        "@types/vscode": "^1.74.0",
+        "@vscode/test-electron": "^2.4.1",
+        "@vscode/vsce": "^3.2.2",
+        "esbuild": "^0.25.0",
+        "eslint": "^9.21.0",
+        "ovsx": "^0.10.1",
+        "proxyquire": "^2.1.3",
+        "ts-node": "^10.9.1",
+        "typescript": "^5.7.3",
+        "typescript-eslint": "^8.25.0",
```
jfeingold35, line 56: Nitpick, but I'd like it if these could stay alphabetized.

src/lib/actions/scanner-action.ts
```diff
@@ -50,7 +50,6 @@ export class ScannerAction {
 		});
 
 		// This has to be a floating promise, because progress bars won't disappear otherwise.
-		// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
jfeingold35, line 53: Is this really not erroring anymore, or did we stop running that rule?

src/test/suite/apex-lsp.test.ts
```diff
@@ -4,13 +4,13 @@
...
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import Sinon = require('sinon');
+import * as Sinon from 'sinon';
 import {expect} from 'chai';
 import * as vscode from 'vscode';
 import { ApexLsp } from '../../lib/apex-lsp';
 
 suite('ScanRunner', () => {
-    let executeCommandStub: sinon.SinonStub;
```
jfeingold35, line 13: Do you have any idea how this was working before? That seems like it should have been a compile error?

src/test/suite/extension.test.ts
```diff
@@ -27,20 +27,19 @@ suite('Extension Test Suite', () => {
...
+			await ext.activate();
 		});
 
 		setup(function () {
 			this.timeout(10000);
 			// Verify that there are no existing diagnostics floating around.
 			const diagnosticsArrays = vscode.languages.getDiagnostics();
 			for (const [uri, diagnostics] of diagnosticsArrays) {
-				expect(diagnostics, `${uri.toString()} should start without diagnostics`).to.be.empty;
+				expect(diagnostics, `${uri.toString()} should start without diagnostics`).to.equal
```
jfeingold35, line 42: Is this really correct? Surely there's an argument we need to pass in?

src/test/suite/fixer.test.ts
```diff
@@ -160,8 +158,7 @@ suite('fixer.ts', () => {
                         test('Should find class start position above the diagnostic line', async () => {
                             const fileUri = vscode.Uri.file(path.join(codeFixturesPath, 'MyClass1.cls'));
 
-                            let doc: vscode.TextDocument;
```
jfeingold35, line 163: This change has me slightly nervous. Do we lose anything by switching to the higher-scoped `doc` declaration here?

src/test/suite/fixer.test.ts
```diff
@@ -185,8 +182,7 @@ suite('fixer.ts', () => {
                         test('Should ignore class defined in single line comment', async () => {
                             const fileUri = vscode.Uri.file(path.join(codeFixturesPath, 'MyClass2.cls'));
 
-                            let doc: vscode.TextDocument;
-                            doc = await vscode.workspace.openTextDocument(fileUri);
+                            await vscode.workspace.openTextDocument(fileUri);
```
jfeingold35, line 185: Does this cause `doc` to be reassigned automatically, or something?

src/test/suite/scanner.test.ts
```diff
@@ -33,6 +34,8 @@ suite('ScanRunner', () => {
 
             // ===== TEST =====
             // Use the scan runner on our target list to create and return our arg array.
+            
+            /* eslint-disable-next-line @typescript-eslint/no-explicit-any */ // TODO: Wow - using "any" here to somehow get access to a private method. Why is this test written this way?
```
jfeingold35, line 38: This is a technique we used for certain tests in the V4 Era, and reused here. It's really helpful when you need your tests to be able to verify something about a private method or property or the like. And while we can look back and say that it's a problem because it tied the tests too closely to the implementation detail, it did seem like a good idea at the time.

src/test/suite/targeting.test.ts
```diff
@@ -79,9 +80,9 @@ suite('targeting.ts', () => {
 
             // ===== ASSERTIONS =====
             // Verify we got the right outputs.
-            expect(targets).to.have.lengthOf(2, 'Wrong number of targets returned');
-            expect(targets[0]).to.equal(path.join(folderPath, "MyClassA1i.cls"), 'Wrong file returned');
-            expect(targets[1]).to.equal(path.join(folderPath, "MyClassA1ii.cls"), 'Wrong file returned');
+            expect(targets).to.have.lengthOf(2, 'Wrong number of targets returned: ' + JSON.stringify(targets, null, 2));
+            expect(targets).to.contain(path.join(folderPath, "MyClassA1i.cls"), 'Wrong file returned');
+            expect(targets).to.contain(path.join(folderPath, "MyClassA1ii.cls"), 'Wrong file returned');
```
jfeingold35, line 85: What necessitated this change?

package.json
stephen-carter-at-sf: They are alphabetized... all except the ones that I am looking to remove... so the list from 58 to 64 I am hoping to remove which is why I put them separately from the others.

src/test/suite/extension.test.ts
stephen-carter-at-sf: Nope - that was a mistake. Thanks!

stephen-carter-at-sf: Fixed.

src/test/suite/targeting.test.ts
stephen-carter-at-sf: Yes... the order should not matter. The order did change when I changed from globby to glob.

src/test/suite/fixer.test.ts
stephen-carter-at-sf: nothing to worry about here  - doc was never used.

src/test/suite/apex-lsp.test.ts
stephen-carter-at-sf: I have no idea... somehow the case didn't matter on the namespace.

jfeingold35 (approved):

PR #186, opened 2025-03-04
CHANGE: @W-17950421@: Add in the ability to run jest unit tests while keeping mocha tests for now

stephen-carter-at-sf:


src/test/legacy/targeting.test.ts
```diff
@@ -31,6 +31,7 @@ suite('targeting.ts', () => {
         test('Given a real file, returns that file', async () => {
             // ===== SETUP =====
             // Get a URI for one file.
+            console.log("HELLO WORLD: " + __dirname)
```
jfeingold35, line 34: Missed console log?

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
 	"compilerOptions": {
+		"skipLibCheck": true, /* needed to avoid conflict between @types/jest and @types/mocha. When we remove mocha, we should remove this line */
```
jfeingold35, line 3: Will we also need to reinstance `strict: false`, or is that unnecessary?

jfeingold35 (changes requested):

src/test/legacy/targeting.test.ts
stephen-carter-at-sf: ah good catch. I was debugging some things. turns out that if you run the tests via the easy button using the "Extension Test Runner" extension, that it doesn't call any compile step first before executing the tests. I'll remove this console log.

tsconfig.json
stephen-carter-at-sf: strict: false is the default. Removing this line is a no-op. I'll add in strict: true in the future when we are ready for it.

PR #187, opened 2025-03-04
CHANGE @W-17526120@ Telemetry for A4D events

jfeingold35:


src/lib/actions/diff-accept-action.ts
```diff
@@ -0,0 +1,41 @@
+import { DiffHunk } from "../../shared/UnifiedDiff";
+import { TelemetryService } from "../core-extension-service";
+import * as Constants from '../constants';
+
+export type DiffAcceptCallback = (diffHunk?: DiffHunk) => Promise<number>;
+
+export type DiffAcceptDependencies = {
+	callback: DiffAcceptCallback;
+	telemetryService: TelemetryService;
+};
```
stephen-carter-at-sf, line 10: Any advantage of having this dependencies type instead of just passing in the telemetry service and callback directly into the constructor?

src/lib/actions/diff-create-action.ts
```diff
@@ -0,0 +1,38 @@
...
+			this.telemetryService.sendCommandEvent(Constants.TELEM_DIFF_SUGGESTION, {
+				commandName: this.commandName,
+				commandSource: 'PLACEHOLDER', // NOT SURE WHAT SHOULD GO HERE.
+			});
+		} catch (e) {
+			const errMsg = e instanceof Error ? e.message : e as string;
+			this.telemetryService.sendException(Constants.TELEM_DIFF_SUGGESTION_FAILED, errMsg, {
+				executedCommand: this.commandName,
+				duration: (Date.now() - startTime).toString()
+			});
```
stephen-carter-at-sf, line 35: Out of curiosity, what does this do? Where does the exception go? 
And do we want to swallow this exception or rethrow it?

src/lib/actions/diff-reject-action.ts
```diff
@@ -0,0 +1,40 @@
...
+		this.telemetryService = dependencies.telemetryService;
+	}
+
+	public async run(diffHunk?: DiffHunk): Promise<void> {
+		const startTime = Date.now();
+		try {
+			await this.callback(diffHunk);
+			this.telemetryService.sendCommandEvent(Constants.TELEM_DIFF_REJECT, {
+				commandName: this.commandName,
+				commandSource: 'PLACEHOLDER', // DECIDE WHAT SHOULD GO HERE
```
stephen-carter-at-sf, line 29: If you look at https://salesforce.quip.com/1ZgwAbICFda1 you'll see example commandSource values from other teams. My guess is we'll want to say something like SalesforceCodeAnalyzer or SFCA or something. Check with Joni Fazo.

Also I wonder how if we are missing anything else on this event. If you haven't done so already look through all the links in that doc which show how other teams are sending in the events.

src/lib/actions/diff-accept-action.ts
jfeingold35: The object keeps the invocation clean and means I don't have to worry about parameters. It also means we can make any of them optional without needing to reorder them or anything like that.

src/lib/actions/diff-create-action.ts
jfeingold35: It sends a telemetry event indicating that we got an exception. I'm not entirely sure where they go, but we've already been doing this style of telemetry for when SFCA fails.
That said, you're almost certainly right that we should be rethrowing instead of swallowing, though in practice it's probably not terribly relevant because the accept/reject functionality should be pretty stable.

src/lib/actions/diff-reject-action.ts
jfeingold35: Yeah, this is one of the things I need to refine.

src/lib/actions/diff-accept-action.ts
```diff
@@ -0,0 +1,41 @@
...
+
+	public async run(diffHunk?: DiffHunk): Promise<void> {
+		const startTime = Date.now();
+		try {
+			const lines: number = await this.callback(diffHunk);
+			this.telemetryService.sendCommandEvent(Constants.TELEM_DIFF_ACCEPT, {
+				commandName: this.commandName,
+				commandSource: 'PLACEHOLDER', // DECIDE WHAT SHOULD GO HERE
+				completionNumLines: lines.toString(),
+				languageType: 'apex' // Apex is the only language A4D codegen supports at present.
```
randi274, line 31: Does CodeAnalyzer provide fixes for other languages? A4D supports all of the Salesforce languages today

jfeingold35: CodeAnalyzer only supports callouts to A4D for a small handful of PMD rules, which are all Apex-based.

src/lib/constants.ts
```diff
@@ -28,6 +28,12 @@ export const TELEM_FAILED_STATIC_ANALYSIS = 'sfdx__codeanalyzer_static_run_faile
 export const TELEM_SUCCESSFUL_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_complete';
 export const TELEM_FAILED_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_failed';
 export const TELEM_SUCCESSFUL_APEX_GURU_FILE_ANALYSIS = 'sfdx__apexguru_file_run_complete';
+export const TELEM_DIFF_SUGGESTION = 'sfdx__eGPT_suggest';
+export const TELEM_DIFF_SUGGESTION_FAILED = 'sfdx__eGPT_suggest_failure';
+export const TELEM_DIFF_ACCEPT = 'sfdx__eGPT_accept';
+export const TELEM_DIFF_ACCEPT_FAILED = 'sfdx__eGPT_accept_failure';
+export const TELEM_DIFF_REJECT = 'sfdx__eGPT_reject';
```
randi274, line 35: I believe the team uses "sfdx__eGPT_clear" to track explicit rejections like this.

src/lib/constants.ts
```diff
@@ -28,6 +28,12 @@ export const TELEM_FAILED_STATIC_ANALYSIS = 'sfdx__codeanalyzer_static_run_faile
 export const TELEM_SUCCESSFUL_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_complete';
 export const TELEM_FAILED_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_failed';
 export const TELEM_SUCCESSFUL_APEX_GURU_FILE_ANALYSIS = 'sfdx__apexguru_file_run_complete';
+export const TELEM_DIFF_SUGGESTION = 'sfdx__eGPT_suggest';
+export const TELEM_DIFF_SUGGESTION_FAILED = 'sfdx__eGPT_suggest_failure';
+export const TELEM_DIFF_ACCEPT = 'sfdx__eGPT_accept';
```
randi274, line 33: I would suggest we also fire out own accept events, as A4D has reasons for monitoring their telemetry, but it would be good to keep our telemetry events consistent as well

src/lib/constants.ts
```diff
@@ -28,6 +28,12 @@ export const TELEM_FAILED_STATIC_ANALYSIS = 'sfdx__codeanalyzer_static_run_faile
 export const TELEM_SUCCESSFUL_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_complete';
 export const TELEM_FAILED_DFA_ANALYSIS = 'sfdx__codeanalyzer_dfa_run_failed';
 export const TELEM_SUCCESSFUL_APEX_GURU_FILE_ANALYSIS = 'sfdx__apexguru_file_run_complete';
+export const TELEM_DIFF_SUGGESTION = 'sfdx__eGPT_suggest';
+export const TELEM_DIFF_SUGGESTION_FAILED = 'sfdx__eGPT_suggest_failure';
```
randi274, line 32: All of the failure states are usually bucketed under "sfdx__eGPT_error" but I'm not sure we need to inform A4D of anything other than accept events. For our own purposes, this would be good to log though!

src/lib/constants.ts
jfeingold35: Okay, I'll switch to that.

src/lib/constants.ts
jfeingold35: Realistically, any of these error cases actually happening is phenomenally low. I'm just programming a little defensively here.

src/lib/actions/diff-accept-action.ts
```diff
@@ -0,0 +1,40 @@
...
+		this.callback = dependencies.callback;
+		this.telemetryService = dependencies.telemetryService;
+	}
+
+	public async run(diffHunk?: DiffHunk): Promise<void> {
+		const startTime = Date.now();
+		try {
+			const lines: number = await this.callback(diffHunk);
+			this.telemetryService.sendCommandEvent(Constants.TELEM_DIFF_ACCEPT, {
+				commandSource: this.commandName,
```
stephen-carter-at-sf, line 28: Can you make this
```
commandSource: 'SFCA_A4D_FIX.' + this.commandName,
```
instead?

Ideally we should be sending the context which initiated the diff from our A4D quick fix button. That is, we should be updating this line:
https://github.com/forcedotcom/sfdx-code-analyzer-vscode/blob/dev/src/modelBasedFixers/apex-pmd-violations-fixer.ts#L86 with the 'SFCA_A4D_FIX' context to be sent to the callback of https://github.com/forcedotcom/sfdx-code-analyzer-vscode/blob/dev/src/extension.ts#L236
and from there you can pass in the 'SFCA_A4D_FIX' string to the DiffCreateAction constructor, and so forth.

src/lib/actions/diff-create-action.ts
```diff
@@ -0,0 +1,38 @@
...
+		this.callback = dependencies.callback;
+		this.telemetryService = dependencies.telemetryService;
+	}
+
+	public async run(code: string, file?: string): Promise<void> {
+		const startTime = Date.now();
+		try {
+			await this.callback(code, file);
+			this.telemetryService.sendCommandEvent(Constants.TELEM_DIFF_SUGGESTION, {
+				commandSource: this.commandName,
```
stephen-carter-at-sf, line 27: ditto

src/lib/actions/diff-reject-action.ts
```diff
@@ -0,0 +1,39 @@
...
+	telemetryService: TelemetryService;
+};
+
+export class DiffRejectAction {
+	private readonly commandName: string;
+	private readonly callback: DiffRejectCallback;
+	private readonly telemetryService: TelemetryService;
+
+	public constructor(commandName: string, dependencies: DiffRejectDependencies) {
+		this.commandName = commandName;
```
stephen-carter-at-sf, line 18: ditto

stephen-carter-at-sf (changes requested):

src/modelBasedFixers/apex-pmd-violations-fixer.ts
```diff
@@ -83,12 +83,12 @@ export class ApexPmdViolationsFixer implements vscode.CodeActionProvider {
 
                 const updatedFileContent = this.replaceCodeInFile(document.getText(), codeSnippet.trim(), diagnostic.range.start.line + 1, diagnostic.range.end.line + 1, document);
                 // Update the command arguments with the resolved code snippet
-                codeAction.command.arguments = [updatedFileContent, document.uri];
+                codeAction.command.arguments = [updatedFileContent, document.uri, Constants.A4D_PREFIX];
```
stephen-carter-at-sf, line 86: Shouldn't Constants.A4D_PREFIX be the first in the array given the callback is:
```
vscode.commands.registerCommand(Constants.UNIFIED_DIFF, async (source: string, code: string, file?: string) => {
```

stephen-carter-at-sf (changes requested):

src/modelBasedFixers/apex-pmd-violations-fixer.ts
jfeingold35: Whoops, yeah.

stephen-carter-at-sf (approved):

PR #190, opened 2025-03-07
FIX: @W-17821243@: Quick fix to external service race condition issues

stephen-carter-at-sf:
Note that this is considered a part of the previous refactoring work to fix a few issues. In the next PR I'll be finishing up all the unit tests and work in this space.


package.json
```diff
@@ -249,14 +249,16 @@
             ],
             "editor/context": [
                 {
-                    "command": "sfca.runOnActiveFile"
+                    "command": "sfca.runOnActiveFile",
+                    "when": "true"
```
randi274, line 253: Was this just a cleanup to make the command explicitly true?

src/extension.ts
```diff
@@ -174,98 +185,100 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
...
+            lines: suggestedCode.split('\n').length.toString()
         });
-    }
+    });
 
 
     // =================================================================================================================
     // ==  Agentforce for Developers Integration and Unified Diff Functionality
     // =================================================================================================================
-    if (Constants.ENABLE_A4D_INTEGRATION && await externalServiceProvider.isLLMServiceAvailable()) {
```
stephen-carter-at-sf, line 209: Note that I also removed the ENABLE_A4D_INTEGRATION check (which is really a part of the next work item). But the main change here is that we would have to wait up to 6 seconds or forcefully activate the A4D extension for this to be true... so instead, I'm passing in a LLMServiceProvider instead to the AgentforceViolationFixer so that it will lazy load the LLMService when it actually needs it. This gives it more time to be available before we attempt to fetch it.

src/lib/agentforce/agentforce-violations-fixer.ts
```diff
@@ -9,47 +9,61 @@ import * as vscode from 'vscode';
...
+        }
+
+        for (const diagnostic of filteredDiagnostics) {
             const fixAction = new vscode.CodeAction(
                 `Fix ${this.extractDiagnosticCode(diagnostic)} using Agentforce.`,
                 vscode.CodeActionKind.QuickFix
             );
-
-            // Pass the diagnostic as an argument to the command
+            fixAction.diagnostics = [diagnostic] // Important (that we used to miss before): we should tie the code fix to the specific diagnostic.
```
stephen-carter-at-sf, line 59: This was missing before. This fixes another small bug that occurs when you have multiple diagnostics in the same range... now the quick fix only applies to the one diagnostic within that range instead of the entire range.

src/extension.ts
```diff
@@ -93,13 +93,28 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     // ==  Code Analyzer Run Functionality
     // =================================================================================================================
     await establishVariableInContext('sfca.codeAnalyzerV4Enabled', () => Promise.resolve(!settingsManager.getCodeAnalyzerV5Enabled()));
+
+    // COMMAND_RUN_ON_ACTIVE_FILE: Invokable by 'commandPalette' and 'editor/context' menu always. Uses v4 instead of v5 when 'sfca.codeAnalyzerV4Enabled'.
```
jfeingold35, line 97: Should this not be `codeAnalyzer.enableV5`, or does it refer to something other than the setting?

src/extension.ts
```diff
@@ -111,41 +126,35 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
...
-            }
-        }
-    });
-
 
     // =================================================================================================================
     // ==  Diagnostic Management Functionality
     // =================================================================================================================
+
+    // TODO: We should look to see if we can make these commands appear conditionally upon whether we have diagnostics
```
jfeingold35, line 134: Pretty sure there's a way to do this. Let's create a work item for looking into it?

src/extension.ts
```diff
@@ -174,98 +185,100 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     // ==  Apex Guru Integration Functionality
     // =================================================================================================================
     const isApexGuruEnabled: () => Promise<boolean> =
-        async () => settingsManager.getApexGuruEnabled() && await ApexGuruFunctions.isApexGuruEnabledInOrg(logger);
+        async () => settingsManager.getApexGuruEnabled() &&
```
jfeingold35, line 188: Looks like something weird is happening with indentation in this file?

package.json
stephen-carter-at-sf: Yeah, trying to be consistent really. The default is "when": "true" but wanted to be explicit to match everything else.

src/lib/constants.ts
```diff
@@ -49,7 +51,6 @@ export const APEX_GURU_MAX_TIMEOUT_SECONDS = 60;
 export const APEX_GURU_RETRY_INTERVAL_MILLIS = 1000;
 
 // A4D Integration
-export const ENABLE_A4D_INTEGRATION = false;
```
randi274, line 52: Will this change then also turn the integration on by default?

src/lib/agentforce/agentforce-violations-fixer.ts
```diff
@@ -62,6 +76,8 @@ export class AgentforceViolationsFixer implements vscode.CodeActionProvider {
         return Constants.A4D_FIX_AVAILABLE_RULES.includes(this.extractDiagnosticCode(diagnostic));
     }
 
+    // TODO: Evaluate if this even should be used. It clearly does more than its responsibility of just resolving the code action... it performs work by invoking the callLLM stuff and updates the existing UNIFIED_DIFF command's arguments before actually executing the unified diff.
+    //       I think we will probably benefit from removing this and instead just have the code action trigger a QF_COMMAND_A4D_FIX which can optionally then trigger a COMMAND_UNIFIED_DIFF or something.
```
randi274, line 80: good callout 👍

src/lib/messages.ts
```diff
@@ -19,6 +19,9 @@ export const messages = {
             increment: 60
         }
     },
+    agentforce: {
+        a4dQuickFixUnavailable: "The ability to fix violations with Agentforce is unavailable since a compatible 'Agentforce for Developers' extension was not found or activated. To enable this functionality, please install the 'Agentforce for Developers' extension and restart VS Code."
```
randi274, line 23: "Compatible" seems to imply a version requirement - do you know if the LLM service cares about a particular version?

src/lib/constants.ts
stephen-carter-at-sf: Yes, see my comment above where I said  I also removed the ENABLE_A4D_INTEGRATION check (which is really a part of the next work item). We plan on doing this this month anyway... and so preemptively I did it here because I moved these checks to inside of the CodeActionProvider anyway. It'll also make things easier to QA.

src/lib/messages.ts
stephen-carter-at-sf: I think the llm service was only made available to other extensions in a specific version of the A4D extension... I don't know exactly which version that is. All this should be somewhere temporary until I get this information from the provider.

src/lib/messages.ts
```diff
@@ -40,21 +43,15 @@ export const messages = {
         existingDfaRunText: "A Salesforce Graph Engine analysis is already running. Cancel it by clicking in the Status Bar.",
     },
     fixer: {
-        supressOnLine: "Suppress violations on this line.",
-        supressOnClass: "Suppress violations on this class.",
+        suppressOnLine: "Suppress violations on this line.",
```
jfeingold35, line 46: Thanks for catching that.

src/extension.ts
stephen-carter-at-sf: I think you put this in originally and I actually agree with it... notice the negation with
```
!settingsManager.getCodeAnalyzerV5Enabled()
```
as it creates the definition of sfca.codeAnalyzerV4Enabled

The reason I agree with it is because soon we'll switch everyone to v5 and remove that getCodeAnalyzerV5Enabled setting in favor of a temporary getCodeAnalyzerV4Enabled setting (which users will opt into in order to stay with v4 for a little while).

But yeah this is all correct as it stands.

src/extension.ts
stephen-carter-at-sf: Not sure what you mean. Looks good to me. The comment I added is only for the 2nd part of the conditional statement.

randi274 (approved):
Looks good!

PR #193, opened 2025-03-17
NEW(a4d): @W-17617362@: Send in context range dependent upon the rule

stephen-carter-at-sf:


src/lib/agentforce/agentforce-violation-fixer.ts
```diff
@@ -34,23 +34,28 @@ export class AgentforceViolationFixer {
...
 
-            // NOTE: We currently assume that the range of the context of the code to replace is equal to the range of
-            //       the violating lines. This won't work for the rules that need additional context, so we have a
-            //       TODO to update the context with W-17617362.
-            const contextRange: vscode.Range = violationLinesRange;
+            const rangeExpander: RangeExpander = new RangeExpander(document);
+            const violationLinesRange: vscode.Range = rangeExpander.expandToCompleteLines(diagnostic.range);
+            let contextRange: vscode.Range = violationLinesRange; // This is the default: ViolationContextScope.ViolationScope
+            if (ruleInfo.violationContextScope === ViolationContextScope.ClassScope) {
+                contextRange = rangeExpander.expandToClass(diagnostic.range);
```
stephen-carter-at-sf, line 48: The forced ClassScope isn't used anywhere currently but is here just in case we decide to use it in the future.

src/lib/logger.ts
```diff
@@ -22,39 +22,36 @@ export class LoggerImpl implements Logger {
     // Displays error message when log level is set to Error, Warning, Info, Debug, or Trace
     error(msg: string): void {
         this.outputChannel.error(msg);
-        this.outputChannel.show(true); // preserveFocus should be true so that we don't make the output window the active TextEditor
```
stephen-carter-at-sf, line 25: I added this in recently, but found that even though we aren't preserving focus... this still causes the outputChannel to show up even when the user may be doing something else.
For example I was looking at the Agentforce for Developers output log and our "Salesforce Code Analyzer" output log came up interrupting what I was looking at.

So now, only debug and trace will show up now.

src/lib/range-expander.ts
```diff
@@ -1,19 +1,96 @@
...
 
-    //Soon we will be adding more methods to this class to expand to the containing method, or the containing class, etc
+    /**
+     * Returns an expanded range to for the entire class that includes the provided range
+     * If not within a class, then we return a range for the whole file.
+     *
+     * @param range the original vscode.Range to be expanded
+     * @return the new expanded vscode.Range
+     */
+    expandToClass(range: vscode.Range): vscode.Range {
```
stephen-carter-at-sf, line 62: I handled a bunch of edge cases that probably will not happen in the real world... but wanted this to be as robust as possible. See the corresponding test file to understand why the algorithm requires us to track inner classes, etc.

.github/workflows/validate-pr.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           # Define regex patterns for different types of PR titles
           MAIN2DEV_REGEX="^MAIN2DEV[[:space:]]*:?[[:space:]]*@W-[[:digit:]]{8,9}@.*MERGING.+[[:digit:]]{1,2}\.[[:digit:]]{1,2}\.[[:digit:]]{1,2}.*"
           RELEASE2MAIN_REGEX="^RELEASE[[:space:]]*:?[[:space:]]*@W-[[:digit:]]{8,9}@.+"
-          PR_INTO_DEV_OR_RELEASE_REGEX="^(FIX|CHANGE|NEW)([[:space:]]*\([a-zA-Z]+\))?[[:space:]]*:?[[:space:]]*@W-[[:digit:]]{8,9}@.+"
+          PR_INTO_DEV_OR_RELEASE_REGEX="^(FIX|CHANGE|NEW)([[:space:]]*\([^)]+\))?[[:space:]]*:?[[:space:]]*@W-[[:digit:]]{8,9}@.+"
```
stephen-carter-at-sf, line 24: I noticed we didn't allow (a4d) in PR title, so I fixed this.

src/lib/agentforce/supported-rules.ts
```diff
@@ -1,31 +1,128 @@
...
+        description: 'Empty Catch Block finds instances where an exception is caught, but nothing is done. In most circumstances, this swallows an exception which should either be acted on or reported.',
+        violationContextScope: ViolationContextScope.MethodScope
+    }],
+    ['EmptyIfStmt', {
+        description: 'Empty If Statement finds instances where a condition is checked but nothing is done about it.',
+        violationContextScope: ViolationContextScope.MethodScope
+    }],
+    ['EmptyStatementBlock', {
+        description: 'Empty block statements serve no purpose and should be removed.',
+        violationContextScope: ViolationContextScope.MethodScope
```
stephen-carter-at-sf, line 55: I updated more than just the 5 rules listed in the work item with MethodScope... there were a few more that seem to need it.

src/test/unit/lib/range-expander.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            const expandedRange: vscode.Range = rangeExpander.expandToMethod(new vscode.Range(5, 9, 5, 20));
+            expect(expandedRange).toEqual(new vscode.Range(5, 0, 5, 25));
+        });
+
+        it('When input range is after but on the same line as the end of a method, then get entire method', () => {
+            const expandedRange: vscode.Range = rangeExpander.expandToMethod(new vscode.Range(4, 9, 4, 13));
+            expect(expandedRange).toEqual(new vscode.Range(2, 0, 4, 21));
+        });
+
+        it('When input range is not on a line with a method, then get the class range', () => {
```
jfeingold35, line 50: Conceptually, it feels weird that a method called `expandToMethod` can actually expand it to a full class instead. Is there a better name we could give it that clarifies this is possible?

src/lib/range-expander.ts
```diff
@@ -1,19 +1,96 @@
...
+        let inInnerClass: boolean = false;
+
+        let startLine: number = range.start.line;
+        while (startLine !== 0 && (!boundaries.isStartOfClass(startLine) || inInnerClass)) {
+            if (boundaries.isEndOfClass(startLine)) {
+                inInnerClass = true;
+            }
+            if (inInnerClass && boundaries.isStartOfClass(startLine)) {
+                inInnerClass = false;
+            }
```
jfeingold35, line 74: I don't know why, but something about this double-`if` rubs me weird.

jfeingold35 (approved):

src/lib/range-expander.ts
stephen-carter-at-sf: It is needed though because a class can start and end on the same line in rare edge cases... so else if wouldn't work here.

src/test/unit/lib/range-expander.test.ts
stephen-carter-at-sf: This is just a fallback. For our use case, the range should always be the code within a method. But I was handling what happens in crazy cases when the diagnostic range doesn't line up anymore with the code because the user types something. So we need this edge case handling until we have automatic updating of our diagnostic ranges.

But I think the name is still appropriate for happy path scenarios.

PR #194, opened 2025-03-20
FIX: @W-18088088@: Fix a few low hanging bugs

stephen-carter-at-sf:
Fixes:
* To avoid us using Trace debug level for now (since it causes some issues with the core extension activating) I switched the Agentforce Fix Diff logging to use debug log instead. Also I made sure it was robust but displaying the original code before we modify the document.
* When the LLM comes back with an explaination, now we show it in an information window instead of just logging it.
* 2 rules needed to switch to ClassScope
* ApexDoc and ExcessiveParameterList are really noisy rules because they highlight the world... making them only hightlight on a single line now
* Updated the class level suppression to show the rule name now so that it doesn't look like have a ton of redundant quick fix suggestions, since they truly are different
* From unified diff, taking out "Ask CodeGenie" in favor of "Agentforce Fix"

### DIFF
--- a/src/lib/diagnostics.ts
+++ b/src/lib/diagnostics.ts
@@ -133,6 +133,11 @@ export class DiagnosticManagerImpl implements DiagnosticManager {
         // This is an interim fix to handle ApexSharingViolations and can be removed once PMD fixes PMD fixes the bug: https://github.com/pmd/pmd/issues/5511
         if (convertible.rule === 'ApexSharingViolations' && primaryLocation.endLine && primaryLocation.endLine !== primaryLocation.startLine) {
             primaryLocationRange = this.convertToRangeForApexSharingViolations(primaryLocation);
+        } else if (convertible.rule === 'ApexDoc' || convertible.rule === 'ExcessiveParameterList') {
+             // See https://github.com/pmd/pmd/issues/5614 and https://github.com/pmd/pmd/issues/5616
+            primaryLocationRange = this.convertToRange(primaryLocation);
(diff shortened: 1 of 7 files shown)


src/lib/diagnostics.ts
```diff
@@ -133,6 +133,11 @@ export class DiagnosticManagerImpl implements DiagnosticManager {
         // This is an interim fix to handle ApexSharingViolations and can be removed once PMD fixes PMD fixes the bug: https://github.com/pmd/pmd/issues/5511
         if (convertible.rule === 'ApexSharingViolations' && primaryLocation.endLine && primaryLocation.endLine !== primaryLocation.startLine) {
             primaryLocationRange = this.convertToRangeForApexSharingViolations(primaryLocation);
+        } else if (convertible.rule === 'ApexDoc' || convertible.rule === 'ExcessiveParameterList') {
```
jfeingold35, line 136: This is a GREAT change, but I'd love if there was a comment explaining why, so whoever's maintaining this the future knows.
```
// These two rules highlight an excessive amount of code, hindering usability. So we shorten them to the first line to keep things readable.`
Or something like that.

**Review by jfeingold35 — APPROVED** [review]

File: src/lib/diagnostics.ts (hunk 1 above)
Comment (stephen-carter-at-sf) [reply src/lib/diagnostics.ts:136]: Not sure if you say I have:
```
// See https://github.com/pmd/pmd/issues/5614 and https://github.com/pmd/pmd/issues/5616
```

PR #202, opened 2025-03-26
CHANGE: @W-18097268@: Refactoring Diagnostics

stephen-carter-at-sf:
In preparation of fixing the diagnostic management, first I simplified a bunch of things with this refactoring work:
* Simplified DiagnosticManager's responsibilities (to just manage diagnostics) and to not call through to expand the targets at all
* Renamed the "DiagnosticConvertible" type to just be "Violation" since that is what it is. Likewise with the CodeLocation.
* Removed the original code and suggested code stuff from the Violation because it doesn't belong there. Instead just directly have apex guru put this on the diagnostic without first putting it on the violation
* Created CodeAnalyzerDiagnostic which extends from vscode.Diagnostic so that we can more easily manage diagnostics
* Simplified the targeting to just do get the full list of impacted files instead of having it also try get the active editor window or error (and instead pushed that responsibility back into the commands themselves).
* And a few smaller updates like:
** Made getting the active text document a bit safer
** Moved some types into their corresponding files instead of having a top level types.ts file.


src/test/unit/lib/diagnostics.test.ts
```diff
@@ -0,0 +1,150 @@
...
+                resources: []
+            };
+
+            const diag: CodeAnalyzerDiagnostic = CodeAnalyzerDiagnostic.fromViolation(violation);
+            expect(diag.range.start.line).toEqual(0);
+        });
+    });
+});
+
+// TODO: Add tests in for the DiagnosticManager
```
stephen-carter-at-sf, line 150: I'll be making some additional changes to the DiagnosticManager in my next PR... so i'll add in all the tests then instead of writing them and reworking them.

src/lib/diagnostics.ts
stephen-carter-at-sf, line 1: This file has the main changes... but github wants to hide the diff. So make sure to load it to see these changes.

src/extension.ts
```diff
@@ -9,7 +9,7 @@
 import * as vscode from 'vscode';
 import {SettingsManager, SettingsManagerImpl} from './lib/settings';
 import * as targeting from './lib/targeting'
-import {DiagnosticManager, DiagnosticManagerImpl} from './lib/diagnostics';
+import {CodeAnalyzerDiagnostic, DiagnosticManager, DiagnosticManagerImpl} from './lib/diagnostics';
```
randi274, line 12: This is interesting to be importing something so feature-specific here - in the future it would make sense to move all of the command logic/registration out as well.

src/lib/diagnostics.ts
```diff
@@ -6,211 +6,185 @@
...
-import * as targeting from './targeting';
-import * as Constants from './constants'
-
-type DiagnosticConvertibleLocation = {
-    file: string;
-    startLine: number;
-    startColumn: number;
+
+export type CodeLocation = {
+    // These all should be optional just like it is over at:
```
randi274, line 11: This is great context! Is this worth keeping as a comment in code, or could we move it out? I think it's helpful today, but not sure it needs to persist past this refactor

src/lib/diagnostics.ts
```diff
@@ -6,211 +6,185 @@
...
-    locations: DiagnosticConvertibleLocation[];
+    locations: CodeLocation[];
     primaryLocationIndex: number;
     resources: string[];
-    currentCode?: string;
-    suggestedCode?: string;
 }
 
+/**
+ * Extended Diagnostic class to hold violation information and uri to make our life easier
```
randi274, line 33: 👏 👏 👏

src/lib/diagnostics.ts
```diff
@@ -6,211 +6,185 @@
...
+ */
+export class CodeAnalyzerDiagnostic extends vscode.Diagnostic {
+    readonly violation: Violation;
+    readonly uri: vscode.Uri;
+
+    private constructor(violation: Violation) {
+        const primaryLocation: CodeLocation = violation.locations[violation.primaryLocationIndex];
+        super(toRange(primaryLocation),
+            messages.diagnostics.messageGenerator(violation.severity, violation.message.trim()),
+            vscode.DiagnosticSeverity.Warning); // TODO: For V5, we should consider using Error for sev 1 and Information for sev 5 instead of always just using Warning.
```
randi274, line 43: 👍

src/lib/apexguru/apex-guru-service.ts
```diff
@@ -118,39 +118,58 @@ export const fileSystem = {
...
+        }],
+        primaryLocationIndex: 0,
+        resources: [
+            'https://help.salesforce.com/s/articleView?id=sf.apexguru_antipatterns.htm&type=5'
+        ]
+    };
+
+    const diagnostic: CodeAnalyzerDiagnostic = CodeAnalyzerDiagnostic.fromViolation(violation);
+    if (suggestedCode.length > 0) {
+        diagnostic.relatedInformation = [
```
randi274, line 161: It looks like this used to live in `convertToDiagnostic` with everything else. This doesn't read quite as cleanly as it used to. Wondering if this could be a separate method, or if this could somehow still live with the rest of the diagnostic logic that's now in `fromViolation` so it's the responsibility of `CodeAnalyzerDiagnostic` to return the complete diagnostic. Thoughts?

src/lib/diagnostics.ts
```diff
@@ -6,211 +6,185 @@
...
-        try {
-            const targets: string[] = await targeting.getTargets(selections);
-
-            for (const target of targets) {
-                this.diagnosticCollection.delete(vscode.Uri.file(target));
-            }
-
-            // TODO: It doesn't make sense that we are performing telemetry here in my opinion. This should be
-            //        moved to an associated action.
-            this.telemetryService.sendCommandEvent(Constants.TELEM_SUCCESSFUL_STATIC_ANALYSIS, {
```
randi274, line 83: We discussed in review that this didn't make sense here, and might in fact have been affecting telemetry data in an unhelpful way, so good to clean up while here 👍

src/lib/fixer.ts
```diff
@@ -22,13 +22,9 @@ export class Fixer implements vscode.CodeActionProvider {
...
         const processedLines = new Set<number>();
-        // Iterate over all diagnostics.
-        return context.diagnostics
-            // Throw out diagnostics that aren't ours, or are for the wrong line.
-            .filter(diagnostic =>
-                diagnostic.source &&
-                diagnostic.source.endsWith(messages.diagnostics.source.suffix)
-                && range.contains(diagnostic.range))
+        // Filter out diagnostics that aren't ours, or are for the wrong line.
+        const codeAnalyzerDiags: CodeAnalyzerDiagnostic[] = context.diagnostics.filter(d => d instanceof CodeAnalyzerDiagnostic);
```
randi274, line 26: Smart check on `instanceof` here!

randi274 (commented):
A good sized change, but does a lot to improve clarity!

src/lib/apexguru/apex-guru-service.ts
stephen-carter-at-sf: I disagree... the fromViolation shouldn't have engine specific variations. Since apexguru (which in the future might be an engine) currently isn't build into code analyzer... i'm trying to keep it separate for now so that all the apex guru refactoring in the near future will not touch all this.

src/lib/agentforce/agentforce-code-action-provider.ts
```diff
@@ -26,13 +26,9 @@ export class AgentforceCodeActionProvider implements vscode.CodeActionProvider {
...
 
         const codeActions: vscode.CodeAction[] = [];
-
-        // Throw out diagnostics that aren't ours, or are for the wrong line.
-        const filteredDiagnostics: vscode.Diagnostic[] = context.diagnostics.filter((diagnostic: vscode.Diagnostic) =>
-            diagnostic.source
-            && diagnostic.source.endsWith(messages.diagnostics.source.suffix)
-            && range.contains(diagnostic.range)
-            && A4D_SUPPORTED_RULES.has(extractRuleName(diagnostic)));
+        const codeAnalyzerDiags: CodeAnalyzerDiagnostic[] = context.diagnostics.filter(d => d instanceof CodeAnalyzerDiagnostic);
```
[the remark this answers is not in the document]
stephen-carter-at-sf: You are right. I thought I couldn't chain because I was first doing a conversion to CodeAnalyzerDiagnostic... but it turns out to either be smart enough or dumb enough that the compiler doesn't need the type conversion first - so as long as the d instanceof CodeAnalyzerDiagnostic comes first. Fixed.

src/lib/apexguru/apex-guru-service.ts
```diff
@@ -27,46 +27,46 @@ export async function isApexGuruEnabledInOrg(logger: Logger): Promise<boolean> {
...
         const errMsg = e instanceof Error ? e.message : e as string;
-        logger.error('Initial Code Analyzer with ApexGuru request failed: ' + errMsg);
+        logger.error('Failed to Scan for Performance Issues with ApexGuru: ' + errMsg);
     }
 }
 
-export function getConvertiblesWithSuggestions(convertibles: DiagnosticConvertible[]): number {
-    // Filter convertibles that have a non-empty suggestedCode and get count
-    return convertibles.filter(convertible => convertible.suggestedCode !== '').length;
+function getDiagnosticsWithSuggestions(diagnostics: CodeAnalyzerDiagnostic[]): CodeAnalyzerDiagnostic[] {
```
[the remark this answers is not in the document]
stephen-carter-at-sf: I actually like having a function be a filter function with a simple single responsibilitiy. The old function was confusing in my opinion. Also, the other benefit is now on lines 56 and 57 there is nice parallelism:

```
                violationCount: diagnostics.length.toString(),
                violationsWithSuggestedCodeCount: getDiagnosticsWithSuggestions(diagnostics).length.toString()
```

Notice how they both now say "length.toString()".

src/lib/apexguru/apex-guru-service.ts
```diff
@@ -118,39 +118,58 @@ export const fileSystem = {
...
+    const encodedCodeAfter = parsed.properties.find((prop: ApexGuruProperty) => prop.name === 'code_after')?.value
+        ?? parsed.properties.find((prop: ApexGuruProperty) => prop.name === 'class_after')?.value
+        ?? '';
+    const suggestedCode: string = Buffer.from(encodedCodeAfter, 'base64').toString('utf8');
+
+    const lineNumber = parseInt(parsed.properties.find((prop: ApexGuruProperty) => prop.name === 'line_number')?.value);
+
+    const violation: Violation = {
+        rule: parsed.type,
+        engine: 'apexguru',
```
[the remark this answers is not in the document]
stephen-carter-at-sf: There ideally shouldn't be any engine specfic code in the VSCode extension at all. 'apexguru' isn't even an engine yet... but because the Violation class is being reused we have to put something in here for now.

And there is discussion that all this code will move to actually be an engine some day. So again - think of this entire file as temporary as we were proving the concept... during planning we'll all be discussing moving apex guru as an engine. I'll add you to those meeting invites so that you can start being a part of those discussions.

src/lib/apexguru/apex-guru-service.ts
```diff
@@ -118,39 +118,58 @@ export const fileSystem = {
...
+        message: parsed.value,
+        severity: 1, // TODO: Should this really be critical level violation? This seems off.
+        locations: [{
+            file: file,
+            startLine: lineNumber,
+            startColumn: 1
+        }],
+        primaryLocationIndex: 0,
+        resources: [
```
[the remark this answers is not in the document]
stephen-carter-at-sf: Again... this was all just a refactor... on pilot code that is going away. So not worth adding in constants at this point.

src/lib/apexguru/apex-guru-service.ts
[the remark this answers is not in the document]
randi274: Got it - that consistency is nice!

randi274 (approved):
Looking great - approved!

PR #208, opened 2025-04-04
FIX: @W-18195076@: Emergency patch for 1.5.x release to fix onOpen issues

stephen-carter-at-sf:
There is a major bug that has existed for quite a while actually that is now showing up due to A4D Extension calling vscode.workspace.openTextDocument even though it doesn't open a file in the editor.

So this change - makes our "Analyze On Open" feature react only to changes to the active text editor. But we don't want to scan each time a user changes focus between tabs, so I added a ScanMonitor to keep track of if a file has been scanned or not yet. Anytime the user saves a file then we remove it from the "alreadyBeenScanned" list so that it can be scanned again if the user re-focuses the window. This was the only approach that seemed to work well and reliably. All other listeners had race conditions and other issues... but this effectively solves the problem.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -67,8 +67,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     const onDidSaveTextDocument = (listener: (e: unknown) => unknown): void => {
         context.subscriptions.push(vscode.workspace.onDidSaveTextDocument(listener));
     }
-    const onDidOpenTextDocument = (listener: (e: unknown) => unknown): void => {
-        context.subscriptions.push(vscode.workspace.onDidOpenTextDocument(listener));
+    const onDidChangeActiveTextEditor = (listener: (e: unknown) => unknown): void => {
+        context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(listener));
     }
 
     // Prepare utilities
@@ -84,6 +84,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     const diagnosticManager: DiagnosticManager = new DiagnosticManagerImpl(diagnosticCollection, telemetryService, logger);
     context.subscriptions.push(diagnosticManager);
     const codeAnalyzerRunner: CodeAnalyzerRunner = new CodeAnalyzerRunner(diagnosticManager, settingsManager, telemetryService, logger);
+    const scanMonitor: ScanMonitor = new ScanMonitor();
+    context.subscriptions.push(scanMonitor);
+
 
     // We need to do this first in case any other services need access to those provided by the core extension.
     // TODO: Soon we should get rid of this CoreExtensionService stuff in favor of putting things inside of the ExternalServiceProvider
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -108,16 +111,25 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
 
         return codeAnalyzerRunner.runAndDisplay(Constants.COMMAND_RUN_ON_ACTIVE_FILE, [document.fileName]);
     });
-    // ... also invoked by opening a file if the user has set things to do so.
-    onDidOpenTextDocument(async (textDocument: vscode.TextDocument) => {
-        if (settingsManager.getAnalyzeOnOpen() && _isValidFileForAnalysis(textDocument.uri)) {
-            await codeAnalyzerRunner.runAndDisplay(Constants.COMMAND_RUN_ON_ACTIVE_FILE, [textDocument.fileName]);
+
+    onDidChangeActiveTextEditor(async (editor: vscode.TextEditor) => {
+        const shouldScan: boolean = editor !== undefined && editor.document.uri.scheme === 'file' && settingsManager.getAnalyzeOnOpen() &&
```
randi274, line 116: nit: can this be broken into separate lines for readability?

stephen-carter-at-sf: What do you suggest?

src/extension.ts
```diff
@@ -376,4 +392,24 @@ async function establishVariableInContext(varUsedInPackageJson: string, getValue
     vscode.workspace.onDidChangeConfiguration(async () => {
         await vscode.commands.executeCommand('setContext', varUsedInPackageJson, await getValueFcn());
     });
+}
+
+class ScanMonitor implements vscode.Disposable {
```
randi274, line 397: This doesn't feel like it should live here - something for a follow-up PR?

stephen-carter-at-sf: Yeah this is all to just get things fixed for now... when we merge into Dev - we'll clean this up. I've already done a number of commits - making this all pretty old code.

src/extension.ts
```diff
@@ -108,16 +111,25 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
...
     });
-    // ... also invoked by opening a file if the user has set things to do so.
-    onDidOpenTextDocument(async (textDocument: vscode.TextDocument) => {
-        if (settingsManager.getAnalyzeOnOpen() && _isValidFileForAnalysis(textDocument.uri)) {
-            await codeAnalyzerRunner.runAndDisplay(Constants.COMMAND_RUN_ON_ACTIVE_FILE, [textDocument.fileName]);
+
+    onDidChangeActiveTextEditor(async (editor: vscode.TextEditor) => {
+        const shouldScan: boolean = editor !== undefined && editor.document.uri.scheme === 'file' && settingsManager.getAnalyzeOnOpen() &&
+            _isValidFileForAnalysis(editor.document.uri) && !scanMonitor.haveAlreadyScannedFile(editor.document.fileName);
+        if (shouldScan) {
```
randi274, line 118: I don't think we had tests for this path before, but could we add them?

stephen-carter-at-sf: The way things are setup right now... I don't think it is trivial to add tests. That's why we desperately need all the refactoring to strip out dependencies to be injected to make testing possible.

randi274 (approved):

src/extension.ts
stephen-carter-at-sf: This this makes short circuiting fail. We end up evaluating more than we need to.

jfeingold35: True, but it looks like the performance trade-off here is pretty minimal, relatively speaking. So maybe it's worth the tiny performance hit to keep the code readable?

stephen-carter-at-sf: OK I'll clean this up when we merge from main to dev... but for now - just trying to get this in.

jfeingold35 (approved):

PR #213, opened 2025-04-08
FIX: @W-18229368@: Do not let partially overlaping diagnostics to swa…

stephen-carter-at-sf:
…llow our quick fix buttons


src/shared/UnifiedDiff.ts
```diff
@@ -483,7 +483,7 @@ export class CodeGenieUnifiedDiffService implements vscode.CodeLensProvider, vsc
         await this.revertUnifiedDiff(diff.document);
 
         if (diff.getHunks().length === 0 || (diff.getHunks().length === 1 && diff.getHunks()[0].type === DiffType.Unmodified)) {
-            vscode.window.showInformationMessage('Agentforce Fix: No changes to diff.');
+            vscode.window.showInformationMessage('No changes to diff.');
```
stephen-carter-at-sf, line 486: Another fix as you go item. We should either say "Agentforce for Developers" or nothing at all. Since this is CodeGenie's unified diff stuff - and may be reused outside of A4D stuff... I thought it was best to just remove this. Note that originally it says "CodeGenie: No changes to diff". So I think this is best.

jfeingold35 (approved):

PR #214, opened 2025-04-08
NEW: @W-18230126@: Prevent our commands from being available until after extension has activated

stephen-carter-at-sf:


src/extension.ts
```diff
@@ -298,6 +300,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     }
 
     telemetryService.sendExtensionActivationEvent(extensionHrStart);
+    vscode.commands.executeCommand('setContext', );
```
AndrewStopchenko-SO, line 303: Please note this - probably a typo

stephen-carter-at-sf: Yup... thanks.

PR #215, opened 2025-04-09
NEW: @W-18235655@: Add in warn/error when editor.CodeLens is not enab…

stephen-carter-at-sf:
…led and A4D Fix is used

### DIFF
--- a/src/lib/messages.ts
+++ b/src/lib/messages.ts
@@ -82,5 +82,9 @@ export const messages = {
         sfdxScannerMissing: "To use the 'Code Analyzer: Use v4 (Deprecated)' setting, you must first install the `@salesforce/sfdx-scanner` Salesforce CLI plugin. But we no longer support v4, so we recommend that you use v5 instead and unselect the 'Code Analyzer: Use v4 (Deprecated)' setting.",
         codeAnalyzerMissing: "To use this extension, first install the `code-analyzer` Salesforce CLI plugin by running `sf plugins install code-analyzer` in the VS Code integrated terminal.",
         coreExtensionServiceUninitialized: "CoreExtensionService.ts didn't initialize. Log a new issue on Salesforce Code Analyzer VS Code extension repo: https://github.com/forcedotcom/sfdx-code-analyzer-vscode/issues"
+    },
+    buttons: {
+        showSettings: 'Show settings',
+        startUsingV5: 'Start using v5'
     }
 };
--- a/src/lib/messages.ts
+++ b/src/lib/messages.ts
@@ -30,7 +30,8 @@ export const messages = {
         noFixSuggested: "No fix was suggested."
     },
     unifiedDiff: {
-        mustAcceptOrRejectDiffFirst: "You must accept/reject the existing diff in this file before performing this action."
+        mustAcceptOrRejectDiffFirst: "You must accept/reject the existing diff in this file before performing this action.",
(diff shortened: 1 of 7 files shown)


src/lib/unified-diff-service.ts
```diff
@@ -41,5 +46,18 @@ export class UnifiedDiffServiceImpl implements UnifiedDiffService {
     async clearDiff(document: vscode.TextDocument): Promise<void> {
         await this.codeGenieUnifiedDiffService.revertUnifiedDiff(document);
     }
+
+    private validateCanShowDiff(): void {
+        if (!this.settingsManager.getEditorCodeLensEnabled()) {
+            const button1Text: string = "Show settings";
```
randi274, line 52: Can this be moved out to a constants file?

randi274 (changes requested):
@stephen-carter-at-sf could you add some telemetry for when the error happens to see how often our users encounter it? My first thought on review was to suggest moving the check earlier, like right before or in [suggest fix](https://github.com/forcedotcom/sfdx-code-analyzer-vscode/blob/6a7111efd894214a8256e17d3acfbe03b1efe3c0/src/lib/agentforce/a4d-fix-action.ts#L37) to save the call to the LLM, but if we can confirm this is something our users don't encounter very often I think this is a tidy way to handle a rare error condition and not worry about increasing LLM calls 👍

stephen-carter-at-sf:
@randi274 We actually do get telemetry for this case. The error gets rethrown, and on line 53 is caught and sent to the telemetry service in handleError.

randi274 (approved):
Connected in Slack and confirmed the re-thrown-error will be captured in telemetry, so I'm good to approve 👍

src/lib/unified-diff-service.ts
stephen-carter-at-sf: Technically it is... its called button1Text.

I don't think this should be in a separate file. If we wanted to share this constant at a level that is used by other methods of this class, then I might move it up higher.

But if you are asking if we should move this to the messages instead... I could do that I suppose. Hold on...

src/lib/messages.ts
```diff
@@ -30,7 +30,8 @@ export const messages = {
         noFixSuggested: "No fix was suggested."
     },
     unifiedDiff: {
-        mustAcceptOrRejectDiffFirst: "You must accept/reject the existing diff in this file before performing this action."
+        mustAcceptOrRejectDiffFirst: "You must accept/reject the existing diff in this file before performing this action.",
```
jshackell-sfdc, line 33:
```suggestion
        mustAcceptOrRejectDiffFirst: "You must accept or reject the existing diff in this file before performing this action.",
```

src/lib/messages.ts
```diff
@@ -30,7 +30,8 @@ export const messages = {
...
+        editorCodeLensMustBeEnabled: "This action requires the 'Editor: Code Lens' setting to be enabled."
```
jshackell-sfdc, line 34:
```suggestion
        editorCodeLensMustBeEnabled: "You must enable the 'Editor: Code Lens' setting before performing this action."
```

src/lib/unified-diff-service.ts
stephen-carter-at-sf: Done.

jshackell-sfdc (changes requested):
See my two suggestions (one is for a message that really isn't part of this PR, but I'd like the little change anyway.)

jfeingold35 (approved):

jfeingold35 (approved):

PR #217, opened 2025-04-10
CHANGE: @W-18246414@ Verify that diff can be shown before calling LLM

stephen-carter-at-sf:
* To respond to recent feedback, I now make it so that A4D Quick Fix doesn't call the LLM until after we have validated that a unified diff can be shown. This saves us from doing unnecessary work.
* Also, I improved a few more things:
** When no fix is suggested from A4D because it returns the exact same code, then we display "No fix suggested" and prevent unified diff tool from doing unnecessary work.
** When a user attempts to edit a file or attempts to perform another A4D quick fix action when an existing diff still exists, then we focus to the diff while showing the warning to make it easy for the user to find it.
** Added in a full test file for our UnifiedDiffServiceImpl layer.

### DIFF
--- a/src/lib/diagnostics.ts
+++ b/src/lib/diagnostics.ts
@@ -70,11 +70,14 @@ export class CodeAnalyzerDiagnostic extends vscode.Diagnostic {
         }
         const diagnostic: CodeAnalyzerDiagnostic = new CodeAnalyzerDiagnostic(violation);
 
-        // Some violation's have ranges that are too noisy, so for now we manually fix them here while we wait on:
-        // - https://github.com/pmd/pmd/issues/5511 for 'ApexSharingViolations'
-        // - https://github.com/pmd/pmd/issues/5614 for 'ApexDoc'
-        // - https://github.com/pmd/pmd/issues/5616 for 'ExcessiveParameterList'
-        if (['ApexDoc', 'ApexSharingViolations', 'ExcessiveParameterList'].includes(violation.rule)) {
+        // Some violations have ranges that are too noisy, so for now we manually fix them here while we wait on PMD to fix them:
+        const rulesToReduceViolationsToSingleLine: string[] = [
+            'ApexDoc',                                      // https://github.com/pmd/pmd/issues/5614
+            'ApexUnitTestMethodShouldHaveIsTestAnnotation', // https://github.com/pmd/pmd/issues/5669
+            'AvoidGlobalModifer',                           // https://github.com/pmd/pmd/issues/5668
+            'ApexSharingViolations',                        // https://github.com/pmd/pmd/issues/5511
+            'ExcessiveParameterList'];                      // https://github.com/pmd/pmd/issues/5616
+        if (rulesToReduceViolationsToSingleLine.includes(violation.rule)) {
             diagnostic.range = new vscode.Range(diagnostic.range.start.line, diagnostic.range.start.character,
                 diagnostic.range.start.line, Number.MAX_SAFE_INTEGER);
         }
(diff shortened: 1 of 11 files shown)


src/lib/diagnostics.ts
```diff
@@ -70,11 +70,14 @@ export class CodeAnalyzerDiagnostic extends vscode.Diagnostic {
         }
         const diagnostic: CodeAnalyzerDiagnostic = new CodeAnalyzerDiagnostic(violation);
 
-        // Some violation's have ranges that are too noisy, so for now we manually fix them here while we wait on:
-        // - https://github.com/pmd/pmd/issues/5511 for 'ApexSharingViolations'
-        // - https://github.com/pmd/pmd/issues/5614 for 'ApexDoc'
-        // - https://github.com/pmd/pmd/issues/5616 for 'ExcessiveParameterList'
-        if (['ApexDoc', 'ApexSharingViolations', 'ExcessiveParameterList'].includes(violation.rule)) {
+        // Some violation's have ranges that are too noisy, so for now we manually fix them here while we wait on PMD to fix them:
```
stephen-carter-at-sf, line 73: Found a few more noisy ranges, so this is just a fix as you go item.

src/lib/diagnostics.ts
```diff
@@ -70,11 +70,14 @@ export class CodeAnalyzerDiagnostic extends vscode.Diagnostic {
...
-        // - https://github.com/pmd/pmd/issues/5614 for 'ApexDoc'
-        // - https://github.com/pmd/pmd/issues/5616 for 'ExcessiveParameterList'
-        if (['ApexDoc', 'ApexSharingViolations', 'ExcessiveParameterList'].includes(violation.rule)) {
+        // Some violations have ranges that are too noisy, so for now we manually fix them here while we wait on PMD to fix them:
+        const rulesToReduceViolationsToSingleLine: string[] = [
+            'ApexDoc',                                      // https://github.com/pmd/pmd/issues/5614
+            'ApexUnitTestMethodShouldHaveIsTestAnnotation', // https://github.com/pmd/pmd/issues/5669
+            'AvoidGlobalModifer',                           // https://github.com/pmd/pmd/issues/5668
+            'ApexSharingViolations',                        // https://github.com/pmd/pmd/issues/5511
+            'ExcessiveParameterList'];                      // https://github.com/pmd/pmd/issues/5616
```
stephen-carter-at-sf, line 79: Found a few more noisy ranges, so this is just a fix as you go item.

src/lib/messages.ts
```diff
@@ -30,7 +30,7 @@ export const messages = {
         noFixSuggested: "No fix was suggested."
     },
     unifiedDiff: {
-        mustAcceptOrRejectDiffFirst: "You must accept/reject the existing diff in this file before performing this action.",
+        mustAcceptOrRejectDiffFirst: "You must accept or reject all changes before performing this action.",
```
stephen-carter-at-sf, line 33: I forgot to implement this feedback from Juliet in the previous PR. We should be using "or" instead of slashes. So doing so here.

src/lib/display.ts
```diff
@@ -35,9 +40,13 @@ export class VSCodeDisplay implements Display {
         this.logger.log(infoMsg);
     }
 
-    displayWarning(warnMsg: string): void {
-        // Not waiting for promise because we didn't add buttons and don't care if user ignores the message.
-        void vscode.window.showWarningMessage(warnMsg);
+    displayWarning(warnMsg: string, ...buttons: DisplayButton[]): void {
```
jfeingold35, line 43: Dumb question, but you've verified what happens when this is an empty array, right?

stephen-carter-at-sf: Yes... a ton of code calls displayWarning without any buttons.

jfeingold35 (approved):

PR #218, opened 2025-04-11
CHANGE: @W-18097339@ Refactor to get CodeAnalyzer class accessible so…

stephen-carter-at-sf:
… that we can soon inject it into A4D Integration to more easily get rule descriptions in the near future.


src/lib/code-analyzer-run-action.ts
```diff
@@ -0,0 +1,114 @@
...
+import * as vscode from "vscode";
+import {messages} from "./messages";
+import {TelemetryService} from "./external-services/telemetry-service";
+import * as Constants from './constants';
+import {CodeAnalyzer} from "./code-analyzer";
+import {Display} from "./display";
+import {getErrorMessage, getErrorMessageWithStack} from "./utils";
+import {ProgressReporter, TaskWithProgressRunner} from "./progress";
+
+export class CodeAnalyzerRunAction {
```
stephen-carter-at-sf, line 13: This file consolidates what used to be the code-analyzer-runner.ts and actions/scanner-action.ts files into a single action. Most of this code was actually just copy/paste

src/lib/code-analyzer.ts
```diff
@@ -0,0 +1,58 @@
...
+    validateEnvironment(): Promise<void>;
+
+    scan(filesToScan: string[]): Promise<Violation[]>;
+
+    getScannerName(): string;
+
+    getRuleDescriptionFor(engineName: string, ruleName: string): Promise<string>;
+}
+
+export class CodeAnalyzerImpl implements CodeAnalyzer {
```
stephen-carter-at-sf, line 16: Instead of having a switch yard inside of the old code-analyzer-runner.ts file... it makes sense to just to use a facade around the two strategies and make a formal CodeAnalyzer interface so that we can start getting closer to what the node api will provide in the future.

src/lib/code-analyzer.ts
```diff
@@ -0,0 +1,58 @@
...
+import {SettingsManager} from "./settings";
+
+export interface CodeAnalyzer {
+    validateEnvironment(): Promise<void>;
+
+    scan(filesToScan: string[]): Promise<Violation[]>;
+
+    getScannerName(): string;
+
+    getRuleDescriptionFor(engineName: string, ruleName: string): Promise<string>;
```
stephen-carter-at-sf, line 13: This currently isn't being used by anything - but soon I'd like to remove our hard coded rule descriptions and so I added this right now to make sure it fits here nicely. Now we should be able to pass around the CodeAnalyzer instance to the other services like our A4D Integration stuff and be able to get the information we need.

src/lib/code-analyzer.ts
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return this.codeAnalyzerV4;
+    }
+
+    private getCodeAnalyzerV5(): CodeAnalyzer {
+        if (!this.codeAnalyzerV5) {
+            this.codeAnalyzerV5 = new CliScannerV5Strategy(this.settingsManager);
+        }
+        return this.codeAnalyzerV5;
+    }
```
stephen-carter-at-sf, line 41: Note that the old code-analyzer-runner.ts file didn't ever cache the strategies... and since they had caching for their validateEnvironment... we weren't ever benefiting from the caching at that layer. Therefore we were actually having to revalidate the environment every single time a scan was performed - which is why things were so slow before in many cases.

Now we only validate once.

src/lib/command-executor.ts
```diff
@@ -0,0 +1,32 @@
+import cp from "node:child_process";
+
+export type CommandOutput = {
+    stdout: string
+    stderr: string
+    exitCode: number
+}
+
+export async function execCommand(command: string, args: string[]): Promise<CommandOutput> {
```
stephen-carter-at-sf, line 9: Until we start using the code analyzer node api... there are a lot of places where we were recreating this promise over and over again. So moving this boiler plate code into a dedicated function.

src/lib/progress.ts
```diff
@@ -0,0 +1,42 @@
...
+    increment?: number;
+};
+
+export interface ProgressReporter {
+    reportProgress(progressEvent: ProgressEvent): void;
+}
+
+export type TaskWithProgress = (progressReporter: ProgressReporter) => Promise<void>;
+
+export interface TaskWithProgressRunner {
```
stephen-carter-at-sf, line 15: This interface is a big win because now we can actually properly unit test our code analyzer run action without requiring vscode.window.withProgress to be in the mix.

src/lib/scanner-strategies/v5-scanner.ts
```diff
@@ -1,105 +1,107 @@
...
+        if (!this.installedCodeAnalyzerVersion) {
             throw new Error(messages.error.codeAnalyzerMissing);
+        } else if (semver.lt(this.installedCodeAnalyzerVersion, absoluteMinVersion)) {
+            throw new Error(
+                messages.error.codeAnalyzerDoesNotMeetMinVersion(
+                    this.installedCodeAnalyzerVersion.toString(),
+                    absoluteMinVersion.toString()
+                ) + '\n' + messages.error.codeAnalyzerMissing);
+        } else if (semver.lt(this.installedCodeAnalyzerVersion, recommendedMinVersion)) {
+            // TODO: Need to somehow throw warning
```
stephen-carter-at-sf, line 62: Once I start leveraging the "--output-file" option of CA5's rules command - in order to get the rule descriptions, then we'll need to be checking the version. So this whole validatePlugin logic really will need to move outside of this class at that time... so that we don't need to attach a display into this class.

Baby steps.

src/lib/progress.ts
```diff
@@ -0,0 +1,42 @@
...
+        this.progressFcn.report(progressEvent);
+    }
+}
+
+export class TaskWithProgressRunnerImpl {
+    async runTask(task: TaskWithProgress): Promise<void> {
+        const progressOptions: ProgressOptions = {
+            location: vscode.ProgressLocation.Notification
+        }
+        const promiseLike: PromiseLike<void> = vscode.window.withProgress(progressOptions, (progressFcn: vscode.Progress<ProgressEvent>): Promise<void> => {
```
jfeingold35, line 36: I'm unfamiliar with PromiseLike. What is that?

stephen-carter-at-sf: vscode often uses their Thenable class which is a PromiseLike. A PromiseLike is like a Promise but without a "catch" implemented. It only has a "then". So I am converting to a Promise here with the Promise.resolve(<promiseLike>)

src/lib/scanner-strategies/v5-scanner.ts
```diff
@@ -1,105 +1,107 @@
...
+    }
+
     public override async scan(filesToScan: string[]): Promise<Violation[]> {
         const resultsJson: ResultsJson = await this.invokeAnalyzer(filesToScan);
 
         return this.processResults(resultsJson);
     }
 
     private async invokeAnalyzer(targets: string[]): Promise<ResultsJson> {
+        const ruleSelector: string = this.settingsManager.getCodeAnalyzerRuleSelectors();
```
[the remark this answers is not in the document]
stephen-carter-at-sf: When we get there, we can easily pass in the user defined selectors instead of the ones on the settings. But we need to keep things dynamic up until we actually do a scan.

jfeingold35 (approved):

PR #219, opened 2025-04-11
CHANGE: @W-18097339@: Remove use of cross-spawn in favor of cp where …

stephen-carter-at-sf:
…we use shell when on windows


src/extension.ts
```diff
@@ -84,9 +84,13 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
 
     const taskWithProgressRunner: TaskWithProgressRunner = new TaskWithProgressRunnerImpl();
 
-    const codeAnalyzer: CodeAnalyzer = new CodeAnalyzerImpl(settingsManager);
+    const codeAnalyzer: CodeAnalyzer = new CodeAnalyzerImpl(settingsManager, display);
     const codeAnalyzerRunAction: CodeAnalyzerRunAction = new CodeAnalyzerRunAction(taskWithProgressRunner, codeAnalyzer, diagnosticManager, telemetryService, logger, display);
 
+    // For performance reasons, it's best to kick this off in the background instead of waiting for a scan to get the
+    // validation out of the way.
+    void codeAnalyzer.validateEnvironment();
```
randi274, line 92: I think this is a good change, and I know it's a small one, but could this be its own PR? I'd like to test this one separately and isolate it as a separate ticket

src/lib/scanner-strategies/v5-scanner.ts
```diff
@@ -28,52 +33,57 @@ type RuleDescription = {
...
-            throw new Error(messages.error.codeAnalyzerMissing);
-        } else if (semver.lt(this.installedCodeAnalyzerVersion, absoluteMinVersion)) {
-            throw new Error(
-                messages.error.codeAnalyzerDoesNotMeetMinVersion(
-                    this.installedCodeAnalyzerVersion.toString(),
-                    absoluteMinVersion.toString()
-                ) + '\n' + messages.error.codeAnalyzerMissing);
-        } else if (semver.lt(this.installedCodeAnalyzerVersion, recommendedMinVersion)) {
-            // TODO: Need to somehow throw warning
+        const installedCliVersion: semver.SemVer | undefined = await this.getInstalledCliVersion();
```
randi274, line 56: Same ask for this - could this move to a subsequent PR for validating the version/environment?

src/lib/settings.ts
```diff
@@ -53,8 +53,13 @@ export class SettingsManagerImpl implements SettingsManager {
         return vscode.workspace.getConfiguration('codeAnalyzer').get('Use v4 (Deprecated)');
     }
 
+    /**
+     * Sets the 'Use v4 (Deprecated)' value at the user (global) level and removes the setting at all other levels
+     */
     public setCodeAnalyzerUseV4Deprecated(value: boolean): void {
         void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', value, vscode.ConfigurationTarget.Global);
+        void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.Workspace);
+        void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.WorkspaceFolder);
```
randi274, line 62: This could also separate to that subsequent PR for testing

randi274 (changes requested):
Overall looking good! @stephen-carter-at-sf could you separate out the changes for the validation timing + fixes for the v4 setting at different levels? Those can all go together, but are quite different than the changes here to fetch rule descriptions for the Quick Fix. I'd like to be able to review and test those changes separately, and I don't think it would hurt to have a WI to capture those changes too.

src/lib/settings.ts
stephen-carter-at-sf: Done. will fix separately with W-18271001

src/extension.ts
stephen-carter-at-sf: Done... will fix separately with W-18271015

randi274 (approved):
Looks good! Thanks for splitting out those changes as well 🥇

PR #220, opened 2025-04-14
CHANGE: @W-18271015@: Attempt to validate that CLI exists immediately…

stephen-carter-at-sf:
… instead of waiting for a scan

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -87,6 +87,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     const codeAnalyzer: CodeAnalyzer = new CodeAnalyzerImpl(settingsManager, display);
     const codeAnalyzerRunAction: CodeAnalyzerRunAction = new CodeAnalyzerRunAction(taskWithProgressRunner, codeAnalyzer, diagnosticManager, telemetryService, logger, display);
 
+    // For performance reasons, it's best to kick this off in the background instead of waiting for a scan to get the
+    // validation out of the way.
+    void codeAnalyzer.validateEnvironment();
+
     // We need to do this first in case any other services need access to those provided by the core extension.
     // TODO: Soon we should get rid of this CoreExtensionService stuff in favor of putting things inside of the ExternalServiceProvider
     await CoreExtensionService.loadDependencies(outputChannel);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -87,6 +87,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     const codeAnalyzer: CodeAnalyzer = new CodeAnalyzerImpl(settingsManager, display);
     const codeAnalyzerRunAction: CodeAnalyzerRunAction = new CodeAnalyzerRunAction(taskWithProgressRunner, codeAnalyzer, diagnosticManager, telemetryService, logger, display);
 
+    // For performance reasons, it's best to kick this off in the background instead of waiting for a scan to get the
+    // validation out of the way.
+    void codeAnalyzer.validateEnvironment();
```
randi274, line 92: Since this is check is now happening upfront, do we still need the check in `code-analyzer-run-action`? We could also await this check and use it conditionally decide whether or not we register any of the commands, which could save some confusion with issues like #222.

stephen-carter-at-sf: There are 2 scenarios to consider:
1) A user opens VS Code and takes a little while before running a scan. In this scenario, we have the time to complete the validation to speed up their first scan.
2) A user immediately opens VS Code, our extension activate, but while things are still validating they want to run a scan. I don't want to prevent them from immediately having the buttons available and so we allow then to start scanning... but the first thing that needs to take place is that it needs to wait for validation to be done. The first one already kicked off - but they were not patient, so we just ask again. Since validation is cached - this only causes issue with immediately first time runs. But I updated the progress prompt to say that it is validating the environment now... so the user sees that something is happening (instead of waiting for it in the background before the button appears).

I think this is the best we can given that there are race conditions with parallel threads. The only alternative is that we would have to add in some sort of polling functions on whether the first one is done validating... before we actually run a scan but that can run into other issues with if the polling never completes - timeouts, etc. So I think this is the best solution.

stephen-carter-at-sf:
@randi274 Regarding your comment (that for some reason it won't allow me to reply to). 
Sure done. Added commit.

src/extension.ts
randi274: Sounds like something where more UI information would help to keep the user updated on that status! Given the desire to want the user to continue if they can, I think this works and in the large majority of cases should help things move faster, as expected 👍

randi274 (approved):

src/extension.ts
stephen-carter-at-sf: Yeah we probably want to put 'SFCA' on the bottom right and have progress updates come through that ideally in the future. Then we could have a check when SFCA is ready on it.

PR #221, opened 2025-04-14
FIX: @W-18271001@: The Enable v5 button should update globally and re…

stephen-carter-at-sf:
…move any local settings


src/test/unit/lib/settings.test.ts
```diff
@@ -0,0 +1,138 @@
+import * as vscode from 'vscode';
+import {SettingsManagerImpl} from "../../../lib/settings";
+
+describe('Tests for the SettingsManagerImpl class ', () => {
```
randi274, line 4: 🎉  Great test coverage!

PR #224, opened 2025-04-16
CHANGE: @W-18238058@: Refactor to remove cross-spawn in favor of CliCommandExecutor interface ...

stephen-carter-at-sf:
and make ScannerStrategy only available post-validation. This way we can more reliably do things within the v5 scanner strategy without having to wonder if it has been validated first.


src/lib/code-analyzer.ts
```diff
@@ -3,60 +3,114 @@ import {CliScannerV4Strategy} from "./scanner-strategies/v4-scanner";
...
+            throw new Error(messages.codeAnalyzer.codeAnalyzerMissing + '\n'
+                + messages.codeAnalyzer.installLatestVersion);
+        } else if (semver.lt(installedVersion, absMinVersion)) {
+            throw new Error(messages.codeAnalyzer.doesNotMeetMinVersion(installedVersion.toString(), recommendedMinVersion.toString()) + '\n'
+                + messages.codeAnalyzer.installLatestVersion);
+        } else if (semver.lt(installedVersion, recommendedMinVersion)) {
+            this.display.displayWarning(messages.codeAnalyzer.usingOlderVersion(installedVersion.toString(), recommendedMinVersion.toString()) + '\n'
+                + messages.codeAnalyzer.installLatestVersion);
+        }
+        this.codeAnalyzerV5 = new CliScannerV5Strategy(installedVersion, this.cliCommandExecutor, this.settingsManager);
```
stephen-carter-at-sf, line 85: Having the installedVersion now as a defined value always on this class is the what I need for the next PR (when I condition some code based on it... now I wont have to see if it is defined anymore before using it).

src/lib/dfa-runner.ts
```diff
@@ -144,7 +146,7 @@ export class DfaRunner implements vscode.Disposable {
                            projectDir: string): Promise<void> {
         const startTime = Date.now();
         try {
-            await verifyPluginInstallation();
+            await this.codeAnalyzer.validateEnvironment(); // Since the ScanRunner currently doesn't take in the codeAnalyzer to run dfa commands, we just validate here
```
stephen-carter-at-sf, line 149: This DfaRunner was doing its own separate validation. Idaelly all this should live underneath the scanner-v4 strategy post validation. But it would cause a much larger refactor. So just keeping the validation here again to maintain existing behavior. Hopefully we will be able to remove all these DFA specific v4 stuff soon enough anyway.

src/lib/scanner-strategies/v4-scanner.ts
```diff
@@ -47,25 +48,22 @@ export type V4ExecutionResult = {
...
+    public constructor(version: semver.SemVer, cliCommandExecutor: CliCommandExecutor, settingsManager: SettingsManager) {
+        this.version = version;
+        this.cliCommandExecutor = cliCommandExecutor;
         this.settingsManager = settingsManager;
     }
 
-    public override getScannerName(): string {
-        return '@salesforce/sfdx-scanner@^4 via CLI';
+    public getScannerName(): Promise<string> {
+        return Promise.resolve(`@salesforce/sfdx-scanner@${this.version.toString()} via CLI`);
```
stephen-carter-at-sf, line 63: The only behavior change is that we have a more precise version number showing up in the logs now.

src/lib/scanner-strategies/v5-scanner.ts
```diff
@@ -72,28 +48,22 @@ export class CliScannerV5Strategy extends CliScannerStrategy {
 
     private async getRuleDescriptionMap(): Promise<Map<string, string>> {
         if (this.ruleDescriptionMap === undefined) {
-            const installedCliVersion: semver.SemVer | undefined = await this.getInstalledCliVersion();
-            if (installedCliVersion && semver.gte(installedCliVersion, '5.0.0-beta.3')) {
-                this.ruleDescriptionMap = await createRuleDescriptionMap();
+            if (semver.gte(this.version, '5.0.0-beta.3')) {
```
stephen-carter-at-sf, line 51: This is one existing example where we no longer need to wonder if the version is defined or not.

src/test/legacy/extension.test.ts
```diff
@@ -217,31 +218,41 @@ suite('Extension Test Suite', () => {
...
             });
 
-            test('Throws error if `sf`/`sfdx` is missing', async function () {
+            test('Throws error if `sf` is missing', async function () {
                 this.timeout(90000);
 
                 // ===== SETUP =====
-                // Simulate SFDX being unavailable.
                 const errorSpy = Sinon.spy(vscode.window, 'showErrorMessage');
-                Sinon.stub(SfCli, 'isSfCliInstalled').resolves(false);
```
stephen-carter-at-sf, line 244: Very soon we will be able to rewrite all these tests as unit tests. But until then, just maintaining things. Here I had to substitute the Sinon stubbing of the old isSfCliInstalled with a stubbed cliCommandExecutor.

src/test/legacy/extension.test.ts
```diff
@@ -328,55 +375,14 @@ suite('Extension Test Suite', () => {
         });
     });
 
-    suite('#verifyPluginInstallation()', () => {
```
stephen-carter-at-sf, line 331: This whole set of tests were tied to a verifyPlugin method that has now moved. Tests for these actually already get covered by the 'Throws ... if ... is missing' tests. So we are safe to remove these tests.

jfeingold35 (approved):

PR #225, opened 2025-04-17
FIX: @W-18271015@: Catching errors appropriately during validation…

stephen-carter-at-sf:
…andd tasks with progress windows


randi274 (dismissed):
Great speedy fix!

PR #226, opened 2025-04-17
NEW: @W-18238058@: Use --target flag for 5.0.0+ and add in lots of tests

stephen-carter-at-sf:


tsconfig.json
```diff
@@ -1,17 +1,17 @@
...
-	"include": [
-		"./src/**/*"
-	]
-}
\ No newline at end of file
+    "compilerOptions": {
+        "skipLibCheck": true, /* needed to avoid conflict between @types/jest and @types/mocha. When we remove mocha, we should remove this line */
+        "module": "NodeNext",
+        "moduleResolution": "NodeNext",
+        "target": "ES2022",
```
stephen-carter-at-sf, line 6: Since our package.json file has
```
    "engines": {
        "vscode": "^1.90.0",
        "node": ">=20.9.0"
    },
```
we are able to go up to ES2022 (probably even ES2023) safely with this version of node. This allows us to use some more modern node functionality (like replaceAll function for example that didn't exist on ES2020). Using ES2022 because that is what we use with core and cli currently.

src/lib/diagnostics.ts
```diff
@@ -26,6 +26,7 @@ export type Violation = {
     severity: number;
     locations: CodeLocation[];
     primaryLocationIndex: number;
+    tags: string[];
```
stephen-carter-at-sf, line 29: Although not consumed yet... I wanted to add this in for completeness so when we read in the violations in v5 - we get everything that we write to the results json file.

src/lib/settings.ts
```diff
@@ -58,8 +58,12 @@ export class SettingsManagerImpl implements SettingsManager {
...
     public setCodeAnalyzerUseV4Deprecated(value: boolean): void {
         void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', value, vscode.ConfigurationTarget.Global);
-        void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.Workspace);
-        void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.WorkspaceFolder);
+
+        // If there is a workspace open (which is true if workspaceFolders is nonempty), then we should update the workspace settings
+        if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
+            void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.Workspace);
+            void vscode.workspace.getConfiguration('codeAnalyzer').update('Use v4 (Deprecated)', undefined, vscode.ConfigurationTarget.WorkspaceFolder);
+        }
```
stephen-carter-at-sf, line 66: I wrestled with tests for a while.

First the tests were failing because I'm now awaiting here. I probably should if we want to in a test file immediately check the value after we set it... but we shouldn't have to wait for vscode to finish these updates from a human timing. And this whole thing will go away at some point, so i'm not worried about these synchronous for testing purposes.

Then the tests were failing with some experiments because we can't update at the workspace or workspace folder level if there is no workspace - instead it actually throws an error in the background. So to fix this we should be actually checking that the workspace exists first.

src/test/legacy/extension.test.ts
```diff
@@ -275,9 +266,9 @@ suite('Extension Test Suite', () => {
 
     suite('#_runAndDisplayDfa()', () => {
         let settingsManager: SettingsManager;
-        
+
         setup(() => {
-            settingsManager = new SettingsManagerImpl();
+            settingsManager = new StubSettingsManager();
```
stephen-carter-at-sf, line 271: It was this test that was giving me trouble and it took me forever to realize it was because of the settings manager not applying my settings immediately. Using a StubSettingsManager which allows me to set values and have them immediately applied without making everything async with awaits... so just using this instead.

src/test/unit/lib/code-analyzer.test.ts
```diff
@@ -0,0 +1,405 @@
...
+                    "-w", "/my/project/dummyFile1.cls",
+                    "-w", "/my/project/dummyFile2.cls",
+                    "-r", "Recommended",
+                    "-f", prePopulatedResultsJsonFile
+                ]);
+
+                expect(violations).toEqual([expectedViolation1, expectedViolation2]);
+            });
+
+            it('When running a scan with version 5.0.0, then confirm we call the cli and process the results correctly using both --workspace and --target', async () => {
```
stephen-carter-at-sf, line 207: Feels good to be able to actually write unit tests now that all the dependencies are injected into the CodeAnalyzer class.

src/test/unit/lib/code-analyzer.test.ts
```diff
@@ -0,0 +1,405 @@
...
+
+            it('The name reflects the currently set v5 version', async () => {
+                cliCommandExecutor.getSfCliPluginVersionReturnValue = new semver.SemVer('5.0.0-beta.3');
+                const scannerName: string = await codeAnalyzer.getScannerName();
+                expect(scannerName).toEqual('code-analyzer@5.0.0-beta.3 via CLI');
+            });
+        });
+
+        describe('v5 tests for the scan method', () => {
+            const sampleResultsJsonText: string = '{\n' +
```
jfeingold35, line 78: Could this be stored in a file?

jfeingold35 (approved):

src/test/unit/lib/code-analyzer.test.ts
stephen-carter-at-sf: Yeah - I was just lazy... let me go ahead and do that real quick.

stephen-carter-at-sf: Done.

PR #227, opened 2025-04-21
NEW: @W-18308504@ Polish first time v5 experience with uninstantiable engines

randi274:
If a user has a setup issue where one or more of their engines can't start during use of the extension, we show them a modified error message directing them to our documentation. We track whether they have seen this message with a key including the date, so that even if their workspace session persists, they will see the message daily to get a nudge to fix the issue and unlock full engine capabilities. 🚗 

Screenshot 2025-04-22 at 10 27 12 AM


src/extension.ts
```diff
@@ -309,7 +312,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
                 settingsManager.setCodeAnalyzerUseV4Deprecated(false);
             } else if (selection === messages.buttons.showSettings) {
                 const settingUri = vscode.Uri.parse('vscode://settings/codeAnalyzer.Use v4 (Deprecated)');
-                vscode.commands.executeCommand('vscode.open', settingUri);
+                vscode.commands.executeCommand(Constants.VSCODE_COMMAND_OPEN_URL, settingUri);
```
randi274, line 315: Since I moved this command to constants, I updated it here too since I could.

src/lib/code-analyzer-run-action.ts
```diff
@@ -112,4 +116,28 @@ export class CodeAnalyzerRunAction {
...
         this.display.displayInfo(messages.info.finishedScan(numFilesScanned, filesWithViolations.size, violations.length));
     }
+
+    /**
+     * An engine won't start, and we want to limit how many times a user has to encounter this error.
+     * If the user has seen the error for this engine in this session - don't show it again.
+     * If it's the first time seeing it, then set the workspace state key and notify the user.
+     */
+    private handleEngineError(engine: string) {
+        // Use a key with date component to reset warnings daily
```
randi274, line 126: I was concerned that the workspace state might persist for longer than we'd like it to. In order to keep this front of mind, I opted for a daily key.

src/lib/code-analyzer-run-action.ts
```diff
@@ -112,4 +116,28 @@ export class CodeAnalyzerRunAction {
...
+        const today = getCurrentDate();
+        const engineWorkspaceKey = `${Constants.ENGINE_WARNING_PREFIX}${engine}_${today}`;
+        const seenThisEngineError = WorkspaceState.getValue<boolean>(engineWorkspaceKey) ?? false;
+
+        if (!seenThisEngineError) {
+            WorkspaceState.setValue(engineWorkspaceKey, true);
```
randi274, line 132: After creating this PR, I'm realizing this could be a good spot for some telemetry. Thoughts?

src/lib/constants.ts
```diff
@@ -42,6 +45,9 @@ export const MINIMUM_REQUIRED_VERSION_CORE_EXTENSION = '58.4.1';
 export const RECOMMENDED_MINIMUM_REQUIRED_CODE_ANALYZER_CLI_PLUGIN_VERSION = '5.0.0';
 export const ABSOLUTE_MINIMUM_REQUIRED_CODE_ANALYZER_CLI_PLUGIN_VERSION = '5.0.0-beta.0';
 
+// workspace state
+export const ENGINE_WARNING_PREFIX = 'engineOff-'
```
randi274, line 49: I am not bound to this const, and happy to rename. I just want something other than "uninstantiable" because that is a hard word to spell. 😅

src/lib/constants.ts
```diff
@@ -56,3 +62,6 @@ export const CONTEXT_VAR_EXTENSION_ACTIVATED = 'sfca.extensionActivated';
 export const CONTEXT_VAR_V4_ENABLED = 'sfca.codeAnalyzerV4Enabled';
 export const CONTEXT_VAR_PARTIAL_RUNS_ENABLED = 'sfca.partialRunsEnabled';
 export const CONTEXT_VAR_APEX_GURU_ENABLED = 'sfca.apexGuruEnabled';
+
+// Documentation URLs
+export const DOCS_PREREQUISITES_LINK = 'https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/setup-cli.html#prerequisites';
```
randi274, line 67: Is this link where we want to point users too? I think so, but want to confirm.

src/lib/messages.ts
```diff
@@ -75,12 +75,14 @@ export const messages = {
     },
     error: {
         analysisFailedGenerator: (reason: string) => `Analysis failed: ${reason}`,
+        engineUninstantiable: (engine: string) => `Heads up! We could not initialize the engine: "${engine}". Click "Learn More" for instructions on how to set up this engine to analyze your code.`,
```
randi274, line 78: @jshackell-sfdc would love a review for tone, and to ensure this message is fully accurate @stephen-carter-at-sf

src/lib/messages.ts
```diff
@@ -75,12 +75,14 @@ export const messages = {
...
         pmdConfigNotFoundGenerator: (file: string) => `PMD custom config file couldn't be located. [${file}]. Check Salesforce Code Analyzer > PMD > Custom Config settings`,
         sfMissing: "To use the Salesforce Code Analyzer extension, first install Salesforce CLI.",
         sfdxScannerMissing: "To use the 'Code Analyzer: Use v4 (Deprecated)' setting, you must first install the `@salesforce/sfdx-scanner` Salesforce CLI plugin. But we no longer support v4, so we recommend that you use v5 instead and unselect the 'Code Analyzer: Use v4 (Deprecated)' setting.",
-        coreExtensionServiceUninitialized: "CoreExtensionService.ts didn't initialize. Log a new issue on Salesforce Code Analyzer VS Code extension repo: https://github.com/forcedotcom/sfdx-code-analyzer-vscode/issues"
+        coreExtensionServiceUninitialized: "CoreExtensionService.ts didn't initialize. Log a new issue on Salesforce Code Analyzer VS Code extension repo: https://github.com/forcedotcom/sfdx-code-analyzer-vscode/issues",
```
randi274, line 82: I'll remove this lagging comma during review feedback 👍

src/lib/constants.ts
jshackell-sfdc:
```suggestion
export const DOCS_PREREQUISITES_LINK = 'https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/analyze-vscode.html#install-and-configure-code-analyzer-vs-code-extension';
```

src/lib/messages.ts
jshackell-sfdc:
```suggestion
        engineUninstantiable: (engine: string) => `Warning: We couldn't initialize the engine: "${engine}". Click "Learn More" for system requirements for this engine, and general instructions on how to set up Code Analyzer.`,
```

jshackell-sfdc: Okay?  We don't have instructions for setting up specific engines, because it's pretty much the same for all engines.  The only difference is the prereq: i.e. Flowtest needs Python, PMD needs Java, and so on.

jshackell-sfdc (changes requested):

src/lib/code-analyzer-run-action.ts
```diff
@@ -112,4 +116,28 @@ export class CodeAnalyzerRunAction {
...
+
+    /**
+     * An engine won't start, and we want to limit how many times a user has to encounter this error.
+     * If the user has seen the error for this engine in this session - don't show it again.
+     * If it's the first time seeing it, then set the workspace state key and notify the user.
+     */
+    private handleEngineError(engine: string) {
+        // Use a key with date component to reset warnings daily
+        const today = getCurrentDate();
+        const engineWorkspaceKey = `${Constants.ENGINE_WARNING_PREFIX}${engine}_${today}`;
```
stephen-carter-at-sf, line 128: As discussed... a few code smells:
* getCurrentDate is accessing global clock. That makes things hard to test without global level mocking. If the class needs a clock, a clock interface should be injected.
* I don't think we need to use the vscode workspaceState to store this state. I think you can just store what you need on this class since this class is tied to the extension when activated.

So either you can simplfy the design and not worry about a clock at all and just have a Map<string,boolean> on the class instance or you can inject a clock and have a Map<string,Date> where you can just check when the last time was when you fired a specific error. The key of the map would be a combination of the error id and the engine it was associated with (basically the building blocks of your error message).

stephen-carter-at-sf (changes requested):

src/lib/constants.ts
stephen-carter-at-sf: @jshackell-sfdc If we point folks over to that https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/analyze-vscode.html#install-and-configure-code-analyzer-vs-code-extension page then can we add to that page a link to show them also how to configure their python or java versions (or disable the engines) by pointing them to the code analyzer config page from there?

src/lib/utils.ts
```diff
@@ -10,6 +10,13 @@ export class RandomUUIDGenerator implements UUIDGenerator {
     }
 }
 
+/**
+ * Gets the current date with format: YYYY-MM-DD
+ */
+export function getCurrentDate(): string {
+    return new Date().toISOString().split('T')[0];
```
stephen-carter-at-sf, line 17: This global state can't be injected.

src/lib/vscode/workspace-state.ts
```diff
@@ -0,0 +1,21 @@
...
+export class WorkspaceState {
+    private static workspaceState: vscode.Memento;
+    
+    public static initialize(context: vscode.ExtensionContext): void {
+        WorkspaceState.workspaceState = context.workspaceState;
+    }
+
+    public static getValue<T>(key: string): T | undefined {
+        return WorkspaceState.workspaceState.get<T>(key);
+    }
```
stephen-carter-at-sf, line 15: This global state can't be injected.

src/test/unit/lib/code-analyzer-run-action.test.ts
```diff
@@ -25,6 +31,9 @@ describe('Tests for CodeAnalyzerRunAction', () => {
         display = new SpyDisplay();
         codeAnalyzerRunAction = new CodeAnalyzerRunAction(taskWithProgressRunner, codeAnalyzer, diagnosticManager,
             telemetryService, logger, display);
+        workspaceGetStub = jest.spyOn(WorkspaceState, 'getValue').mockReturnValue(false);
+        workspaceSetStub = jest.spyOn(WorkspaceState, 'setValue').mockReturnValue(undefined);
+        jest.spyOn(utils, 'getCurrentDate').mockReturnValue('YYYY-MM-DD');
```
stephen-carter-at-sf, line 36: Unless we are just testing a concrete implementation of one of our interfaces (i.e. one of our <NameOfInterface>Impl) classes, then we shouldn't be using global level mocking (like with jest.spyOn) but instead should be injecting dependencies, and creating spies and stubs manually.

src/lib/constants.ts
jshackell-sfdc: @stephen-carter-at-sf , the last step in the "Install and Configure" tasks (for both local VSCode and Code Builder) says that you can optionally configure code analyzer by creating a YAML file, and then points users here: 
 https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/config-custom.html.  I thought that was enough, but let me know if you still think I need to make it more obvious.

src/lib/messages.ts
randi274: That's much clearer - thank you!

src/lib/code-analyzer-run-action.ts
```diff
@@ -112,4 +116,26 @@ export class CodeAnalyzerRunAction {
...
         this.display.displayInfo(messages.info.finishedScan(numFilesScanned, filesWithViolations.size, violations.length));
     }
+
+    /**
+     * An engine won't start, and we want to limit how many times a user has to encounter this error.
+     * If the user has seen the error for this engine in this session - don't show it again.
+     * If it's the first time seeing it, then store that and notify the user.
+     */
+    private handleEngineError(engine: string) {
+        const engineWorkspaceKey = `${Constants.ENGINE_WARNING_PREFIX}${engine}`;
```
stephen-carter-at-sf, line 126: Instead of using the ENGINE_WARNING_PREFIX... you could just use `UNINSTANTIABLE_ENGINE_RULE` as the prefix. Also maybe don't call it "WorkspaceKey" but "ErrorKey":

```
const errorKey = `${UNINSTANTIABLE_ENGINE_RULE}-${engine}`;
```

src/test/unit/lib/code-analyzer-run-action.test.ts
```diff
@@ -73,3 +109,16 @@ function createSampleViolation(suffix: string, severityLevel: number, locations:
         resources: []
     };
 }
+
+function createUninstantiableViolation(suffix: string, severityLevel: number, locations: CodeLocation[]): Violation {
+    return {
+        rule: UNINSTANTIABLE_ENGINE_RULE,
+        engine: `${suffix}`,
```
stephen-carter-at-sf, line 116: Since this no longer is a suffix, can we rename the variable to be something like `engineName` instead of suffix.

src/test/unit/lib/code-analyzer-run-action.test.ts
```diff
@@ -73,3 +109,16 @@ function createSampleViolation(suffix: string, severityLevel: number, locations:
         resources: []
     };
 }
+
+function createUninstantiableViolation(suffix: string, severityLevel: number, locations: CodeLocation[]): Violation {
```
stephen-carter-at-sf, line 113: You don't need to pass in the severityLevel or locations since you always seem to pass in the same ones. Just hard code them in the return value.

src/lib/code-analyzer-run-action.ts
randi274: Works for me!

src/lib/code-analyzer.ts
```diff
@@ -11,7 +11,7 @@ import {
     RECOMMENDED_MINIMUM_REQUIRED_CODE_ANALYZER_CLI_PLUGIN_VERSION
 } from "./constants";
 import {CliScannerStrategy} from "./scanner-strategies/scanner-strategy";
-import {VscodeWorkspace, VscodeWorkspaceImpl} from "./vscode-api";
+import {VscodeWorkspace, VscodeWorkspaceImpl} from "./vscode/vscode-api";
```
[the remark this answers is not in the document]
randi274: Agreed - I was on the fence about putting it back. Is this helpful in moving us in the direction you're hoping to go, or would you rather I undo (wrong direction/keep it clearer, etc)?

PR #235, opened 2025-04-30
FIX: @W-18403733@: Pass in raw selection to be scanned instead of expanding…

stephen-carter-at-sf:
… to avoid passing too many args to the cli


src/test/unit/lib/code-analyzer.test.ts
```diff
@@ -139,6 +145,7 @@ describe('Tests for the CodeAnalyzerImpl class', () => {
                     "code-analyzer", "run",
                     "-w", "/my/project/dummyFile1.cls",
                     "-w", "/my/project/dummyFile2.cls",
+                    "-w", "/my/project/subfolder", // Should not be expanded to files - should stay as raw folder
```
stephen-carter-at-sf, line 148: This is the test for the fix.

randi274 (approved):

PR #246, opened 2025-06-10
FIX: @W-18671146@ Update Service Provider Initialization

randi274:
Update for telemetry - extension name and reporter updates.


src/lib/external-services/external-service-provider.ts
```diff
@@ -87,7 +89,8 @@ export class ExternalServiceProvider implements LLMServiceProvider, TelemetrySer
         }
 
         try {
-            const coreTelemetryService: TelemetryServiceInterface = await ServiceProvider.getService(ServiceType.Telemetry, Constants.EXTENSION_ID);
+            const coreTelemetryService: TelemetryServiceInterface = await ServiceProvider.getService(ServiceType.Telemetry);
+            await coreTelemetryService.initializeService(this.extensionContext);
```
randi274, line 93: After testing a few different rounds, I think this was the key missing piece. Adding the name into `getService` was causing our telemetry to not be sent.

stephen-carter-at-sf (approved):

PR #267, opened 2025-08-07
FIX: @W-19160212@: Fix argument order of A4D QF action callback

stephen-carter-at-sf:


src/extension.ts
```diff
@@ -261,7 +261,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<SFCAEx
     const a4dFixAction: A4DFixAction = new A4DFixAction(externalServiceProvider, codeAnalyzer, unifiedDiffService, 
         diagnosticManager, telemetryService, logger, display);
     const a4dFixActionProvider: A4DFixActionProvider = new A4DFixActionProvider(externalServiceProvider, logger);
-    registerCommand(A4DFixAction.COMMAND, async (document: vscode.TextDocument, diagnostic: CodeAnalyzerDiagnostic) => {
```
stephen-carter-at-sf, line 264: Great example of how unit testing can't catch everything. So in the future when I separate out the legacy tests into their own end-to-end module... we'll want to test a workflow like this from the UI if possible.

PR #283, opened 2025-08-29
NEW: @W-18538308@: Make ApexGuru integration GA by removing pilot flag...

stephen-carter-at-sf:
… and respond to org changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
         "jest": "^30.0.5",
         "jest-mock-vscode": "^4.6.0",
         "mocha": "^10.8.2",
+        "npm-run-all": "^4.1.5",
         "ovsx": "^0.10.5",
         "proxyquire": "^2.1.3",
         "rimraf": "*",
@@ -140,11 +141,6 @@
                         "type": "boolean",
                         "default": false,
                         "description": "Scan files on open automatically."
-                    },
-                    "codeAnalyzer.apexGuru.enabled": {
-                        "type": "boolean",
-                        "default": false,
-                        "description": "(Pilot) Discover critical problems and performance issues in your Apex code with ApexGuru, which analyzes your Apex files for you. This feature is in a closed pilot; contact your account representative to learn more."
                     }
                 }
             },
@@ -188,7 +184,7 @@
                 },
                 {
                     "command": "sfca.runApexGuruAnalysisOnCurrentFile",
-                    "when": "sfca.extensionActivated && sfca.apexGuruEnabled && resourceExtname =~ /\\.cls|\\.trigger|\\.apex/"
+                    "when": "sfca.extensionActivated && sfca.shouldShowApexGuruButtons && resourceExtname =~ /\\.cls|\\.trigger|\\.apex/"
                 }
             ],
             "editor/context": [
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
         "jest": "^30.0.5",
         "jest-mock-vscode": "^4.6.0",
         "mocha": "^10.8.2",
+        "npm-run-all": "^4.1.5",
```
stephen-carter-at-sf, line 63: This was a missing dependency preventing our `npm run watch` (indirectly by the .vscode > launch.json and tasks.json files) from working.

src/test/unit/test-utils.ts
```diff
@@ -28,4 +29,31 @@ export function createSampleViolation(location: CodeLocation, ruleName: string =
         fixes: fixes,
         suggestions: suggestions
     };
+}
+
+// To help test asyncronous code (when we purposely do not await a promise), we can use this function to help wait for
+// async operation to take place successfully. If it doesn't within the specified timeout, then an exception is thrown.
+export async function expectEventuallyIsTrue(conditionFn: () => boolean, timeout: number = 5000, interval: number = 50): Promise<void> {
```
jfeingold35, line 36: This seems like a useful utility, but is it redundant with `node:util`'s `promisify`?

package.json
```diff
@@ -140,11 +141,6 @@
                         "type": "boolean",
                         "default": false,
                         "description": "Scan files on open automatically."
-                    },
-                    "codeAnalyzer.apexGuru.enabled": {
```
randi274, line 144: 🎉

package.json
```diff
@@ -188,7 +184,7 @@
                 },
                 {
                     "command": "sfca.runApexGuruAnalysisOnCurrentFile",
-                    "when": "sfca.extensionActivated && sfca.apexGuruEnabled && resourceExtname =~ /\\.cls|\\.trigger|\\.apex/"
+                    "when": "sfca.extensionActivated && sfca.shouldShowApexGuruButtons && resourceExtname =~ /\\.cls|\\.trigger|\\.apex/"
```
randi274, line 187: Nice! I like this as an upgraded context variable name.

src/lib/apexguru/apex-guru-service.ts
```diff
@@ -67,42 +73,60 @@ export class LiveApexGuruService implements ApexGuruService {
...
+            // the user to even have access to any of the ApexGuru scan buttons. If it does, we should investigate.
+            throw new Error('The getAvailability method should not be called until updateAvailability is first called');
         }
         return this.availability;
     }
 
-    // TODO: Soon with W-18538308 we will be using the connection.onOrgChange to wire up to this
-    private async updateAvailability(): Promise<void> {
+    onAccessChange(callback: (access: ApexGuruAccess) => void): void {
+        this.eventEmitter.addListener(ACCESS_CHANGED_EVENT, callback);
```
randi274, line 93: Is it worth adding a cleanup for the eventEmitter to prevent memory leaks? I know we're not doing this with our other event emitters (so we could do that in a future ticket as well)

stephen-carter-at-sf: Not sure what you mean. The EventEmitter is tied to the LiveApexGuruService which is tied to the Extension. I could be wrong, but I believe that if the extension is deactivated, then the LiveApexGuruService is destroyed and thus the EventEmitter is destroyed. Also there is only one registered subscriber for the event and so I'm not sure how memory leaks could occur. Can you elaborate?

stephen-carter-at-sf: The only mention of where memory leaks occur is if many listeners are registered. But we only have one. And note that the default limit is 10. So again, I'm not sure what you mean.
image

jfeingold35 (approved):

src/lib/apexguru/apex-guru-service.ts
randi274: Cool, I think that looks good, given the limit!

randi274 (approved):

PR #341, opened 2026-03-17
NEW @W-20485780@ - Add Fixes and Suggestions support 

nikhil-mittal-165:


aruntyagiTutu (approved):

src/lib/fix-suggestion.ts
```diff
@@ -8,7 +8,6 @@ export type CodeFixData = {
     diagnostic: vscode.Diagnostic
 
     // The range of the original context within the document that is suggested to be replaced with a code fix
-    // IMPORTANT: It is assumed that this range includes the entire start and end lines and not partial
```
namrata111f, line 11: Why are we removing this comment? From the below usage it looks to me we are assuming the range is always going have both start and end lines.

src/lib/fix-suggestion.ts
```diff
@@ -69,14 +68,29 @@ export class FixSuggestion {
     }
 
     getFixedDocumentCode(): string {
```
namrata111f, line 70: Why do we need changes here? Shouldn't the get fix keep working out of the box?

src/lib/code-analyzer.ts
```diff
@@ -131,6 +139,16 @@ export class CodeAnalyzerImpl implements CodeAnalyzer {
             args.push('-c', configFile);
         }
 
+        // Add flags only if CLI supports them (>= 5.12.0) and settings are enabled
+        if (this.supportsFixesAndSuggestions) {
+            if (this.settingsManager.getIncludeFixes()) {
+                args.push('--include-fixes');
+            }
+            if (this.settingsManager.getIncludeSuggestions()) {
+                args.push('--include-suggestions');
```
namrata111f, line 148: Are we also showing these settings to the customer if the cli is 5.12 or above?

nikhil-mittal-165: yes settings are always shown to user for greater control

